PR #164, opened 2022-10-02
Add Python bindings

recmo:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


recmo:
We have `pyo3` conversion to/from native integers now. This should cover most needs. Might revisit if we have interesting wrappers like a `PrimeField` type.

PR #171, opened 2022-10-18
feat(support): Parity SCALE Codec

shekhirin:
## Motivation

Implement SCALE codec for `Uint` along with the compact types.

## Solution

Most of the compact encoding implementation is from https://github.com/paritytech/parity-scale-codec/blob/ca03f7fd981855b7cd6348fe32a9831e8e142813/src/compact.rs, but adapted to work with bits number other than power of 2 and up to 536 (according to the spec) instead of 128.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


shekhirin:
@recmo fyi it's ready for review

src/support/scale.rs
```diff
@@ -0,0 +1,319 @@
...
+
+impl<const BITS: usize, const LIMBS: usize> Decode for Uint<BITS, LIMBS> {
+    fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+        Decode::decode(input).and_then(|b: Vec<_>| {
+            Self::try_from_le_slice(&b).ok_or(Error::from("value is larger than fits the Uint"))
+        })
+    }
+}
+
+// TODO: Use nightly generic const expressions to validate that BITS parameter is less than 536
```
recmo, line 40: A workaround is to add `assert!(BITS < 536)`s in the constructs. Since this is const it will optimize away if everything is good. If the users makes a mistake, it will panic the moment a type is constructed. Presumably the users will at least construct the type during testing.

recmo: Ah, you have this with `assert_compact_supported`, I would add it to the `From` construtors too.

recmo (approved):
Looks great! Thanks for the contribution!

PR #190, opened 2022-10-25
feat(support): hash types for `primitive-types`

shekhirin:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


src/aliases.rs
```diff
@@ -49,5 +49,17 @@ pub type U2048 = Uint<2048, 32>;
...
+pub type H128 = Bits<128, 2>;
+
+/// [`Bits`] for `160` bits.
+pub type H160 = Bits<160, 3>;
+
+/// [`Bits`] for `256` bits.
+pub type H256 = Bits<256, 4>;
+
+/// [`Bits`] for `512` bits.
+pub type H512 = Bits<512, 8>;
```
recmo, line 62: I don't like the `H` prefix because these are bit arrays, they have little to do with hashes.

src/aliases.rs
```diff
@@ -49,5 +49,17 @@ pub type U2048 = Uint<2048, 32>;
...
+
 // TODO: B0, B1, B8, ... B4096
```
recmo, line 64: We  should implement all of these. probably using a macro

src/bit_arr.rs
```diff
@@ -68,13 +69,13 @@ macro_rules! forward {
             }
         )*
     };
-    ($(fn $fnname:ident(&self) -> $res:ty;)*) => {
+    ($(fn $fnname:ident$(<$(const $generic_arg:ident:$generic_ty:ty),+>)?(&self) -> $res:ty;)*) => {
```
recmo, line 72: This is some great macro fu!

src/aliases.rs
shekhirin: oops, turns out that's what `B` is about 🫡

gakonst:
so sick

PR #200, opened 2022-10-31
Check for overflows when parsing uints from macro 

### LINKED ISSUE #199: Instantiating a specific integer results in an overflow error, even though it clearly fits within the specified size
**Version**
1.6

**Platform**
Darwin computer.local 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:28:41 PST 2021; root:xnu-8019.61.5~1/RELEASE_ARM64_T6000 x86_64

**Description**
Attempting to instantiate the number `258664426012969093929703085429980814127835149614277183275038967946009968870203535512256352201271898244626862047232` as a `Uint<384, 6>` results in overflow errors at compile time.

The following code:

```rust
use ruint::{uint, Uint};

fn main() {
    println!("{}", Uint::<384, 6>::MAX);
    let _a = uint!(258664426012969093929703085429980814127835149614277183275038967946009968870203535512256352201271898244626862047232_U384);
}
```

Resulted in this error:

```
jules@computer test-uint % cargo run
   Compiling test-uint v0.1.0 (/Users/jules/workspace/test-uint)
error: proc macro panicked
 --> src/main.rs:5:14
  |
5 |     let _a = uint!(258664426012969093929703085429980814127835149614277183275038967946009968870203535512256352201271898244626862047232_U384);
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: attempt to add with overflow

error: could not compile `test-uint` due to previous error
```

When putting the `MAX` and the given integer next to each other:
`39402006196394479212279040100143613805079739270465446667948293404245721771497210611414266254884915640806627990306815`

`258664426012969093929703085429980814127835149614277183275038967946009968870203535512256352201271898244626862047232`

It's clear that the integer should at the very least fit within the bounds of 384 bits.

So far from simple checks, the compiler will allow it if the number is decremented by 1, but still error if it is incremented by 1. I have not checked any other values for now, though I am capable of successfully instantiating `Uint`s of the same bitsize with much higher values, without any issue.

julesdesmit:
## Motivation

Certain integers when parsed with the `uint!` macro can cause an overflow. This PR should introduce a simple fix for this edge case.

## Solution

When the limbs are being constructed, digits are now added to the smallest limb by using `overflowing_add`, after which we check for overflow. If overflow has occurred, the next limb is incremented by 1.

This PR should fix #199 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation - N/A
- [x] Updated the changelog


ruint-macro/src/lib.rs
```diff
@@ -110,7 +110,11 @@ fn parse(value: &str, bits: &str) -> Result<TokenStream, String> {
         }
 
         // Add digit to result
-        limbs[0] += digit; // Never carries
+        let (result, overflow) = limbs[0].overflowing_add(digit);
+        limbs[0] = result;
+        if overflow {
+            limbs[1] += 1;
+        }
```
recmo, line 117: This is not sufficient, we need to carry propagate the whole way.

ruint-macro/src/lib.rs
```diff
@@ -198,4 +202,9 @@ mod tests {
             assert_eq!(0o10_U8, 8_u64.try_into().unwrap());
         }
     }
+
+    #[test]
+    fn test_overflow_during_parsing() {
+        let _ = uint!(258664426012969093929703085429980814127835149614277183275038967946009968870203535512256352201271898244626862047232_U384);
```
recmo, line 208: Using `uint!` in tests is actually a bit cyclical and problematic. I should change that.

recmo (commented):
This is awesome! Thanks for finding and diagnosing this issue! I'll take it from here.

ruint-macro/src/lib.rs
```diff
@@ -91,13 +91,12 @@ fn parse(value: &str, bits: &str) -> Result<TokenStream, String> {
...
-                "Invalid digit {} in base {} (did you forget the `0x` prefix?)",
-                c, base
+                "Invalid digit {c} in base {base} (did you forget the `0x` prefix?)"
             ));
         }
 
-        // Multiply result by base
-        let mut carry = 0_u64;
+        // Multiply result by base and add digit
+        let mut carry = digit;
```
recmo, line 99: Turns out we can trivially merge it with the basis multiplication :)

julesdesmit: That's amazing! So much cleaner :)

PR #202, opened 2022-11-01
feat(macro): Bits construction

shekhirin:
## Motivation

Support constructing `Bits` type in `uint!` macro.

## Solution

Implement it.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/ruint-macro/src/lib.rs
+++ b/ruint-macro/src/lib.rs
@@ -14,10 +14,12 @@ enum LiteralBaseType {
 }
 
 impl LiteralBaseType {
-    const fn delimiter(self) -> &'static str {
-        match self {
-            Self::Uint => "U",
-            Self::Bits => "B",
+    fn delimiter(self, source: &str) -> &'static str {
+        let prefix = source.split_at(2).0;
+        match (self, prefix) {
+            (Self::Uint, _) => "U",
+            (Self::Bits, "0x") => "_B",
+            (Self::Bits, _) => "B",
         }
     }
 }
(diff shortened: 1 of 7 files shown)


ruint-macro/src/lib.rs
```diff
@@ -120,19 +147,19 @@ fn parse(value: &str, bits: &str) -> Result<(usize, Vec<u64>), String> {
...
     Ok((bits, limbs))
 }
 
 /// Transforms a [`Literal`] and returns the substitute [`TokenTree`]
-fn transform_literal(literal: Literal) -> TokenTree {
+fn transform_literal(literal: Literal, base_type: LiteralBaseType) -> TokenTree {
     let source = literal.to_string();
-    if let Some((value, bits)) = source.split_once('U') {
-        let tokens = parse(value, bits).map_or_else(
+    if let Some((value, bits)) = source.split_once(base_type.delimiter()) {
```
recmo, line 159: We should read the `base_type` from the literal, so you can freely mix `U256` and `B256` (and soon `I256`) in the contents.

recmo, line 159: When `base_type` is `B` this will split on the first `B`, which may be a valid hex digit! We should use `rsplit_once` instead to start searching from the back. But even then there is potential for confusion:
`0x90BB13` could be a valid regular rust int literal, or a `B13` literal. Maybe we should require a `_` before the suffix? We can also make it a requirement just for the (hex, bits) combination.

https://doc.rust-lang.org/reference/tokens.html#integer-literals

ruint-macro/src/lib.rs
```diff
@@ -162,15 +189,25 @@ fn transform_tree(tree: TokenTree) -> TokenTree {
...
 #[proc_macro]
 pub fn uint(stream: TokenStream) -> TokenStream {
-    transform_stream(stream)
+    transform_stream(stream, LiteralBaseType::Uint)
+}
+
+// Repeat the crate doc
+#[doc = include_str!("../Readme.md")]
+#[proc_macro]
+pub fn bits(stream: TokenStream) -> TokenStream {
```
recmo, line 209: One of the nice things about the `uint!` macro is that you can do this:

```
uint!{

let a = 42_U256;
let b = 0xf00f_1337_c0d3_U256;
let c = a + b;
assert_eq!(c, 263947537596669_U256);

}
```

I'd like to keep that behaviour and extend it with support for `B` so you can mix both:

```
uint!{

let foo = 0xf0f1337_U32;
let goo = 0x90bb1e_B32;

}
```

recmo (changes requested):

ruint-macro/src/lib.rs
shekhirin: > Maybe we should require a _ before the suffix?
yep, that sounds good.

PR #225, opened 2023-02-22
skip leading zeros when serializing uint

### LINKED ISSUE #224: Uint Serialize impl does not skip leading zeros
**Version**
List the versions of the service you are using. The easiest way to get
this information is using the `cargo tree` subcommand:

ruint v1.7.0

**Description**

The Serialize impl of `Uint` is not compliant with the ethereum rpc quantity format: https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding

I tried this code:

```rust

let val = serde_json::to_string(&U256::ZERO).unwrap();
// `0x0000000000000000000000000000000000000000000000000000000000000000`
```

I expected to see this happen:

quantity format requires this to be `0x0`

Ideally the Serialize impl of `Uint` would skip leading zeros. This is also how primitive-types::U256 is serialized: https://github.com/paritytech/parity-common/blob/master/primitive-types/impls/serde/src/serialize.rs#L21

mattsse:
## Motivation
Attempts to close #224

Skip leading zeros for `Uint` for human readable serializer.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/support/serde.rs
+++ b/src/support/serde.rs
@@ -13,22 +13,12 @@ use std::{fmt::Write, str};
 /// Serialize a [`Uint`] value.
 ///
 /// For human readable formats a `0x` prefixed lower case hex string is used.
-/// For binary formats a byte array is used. Leading zeros are included.
+/// For binary formats a byte array is used.
+/// Leading zeros are skipped in human-readable formats. If the uint consists
+/// entirely of zeros, `0x0` is serialized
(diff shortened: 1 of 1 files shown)


prestwich:
this needs to be rebased on latest main due to changes in e35abc879492d9857f7b1039b77d62d613ca158d

prestwich:
per convo with @mattsse, I'm closing this in favor of #243

PR #229, opened 2023-03-30
fix: disallow deser for strings that are too long

Rjected:
## Motivation

https://github.com/paradigmxyz/reth/issues/1784 highlighted that deserialization allows strings that are too long, for example `0x000` would be successfully deserialized for a single-byte `Uint`.

## Solution

Ensure that the input string to deserialize is not longer than 2 times the max number of bytes.

Added tests for max value (`0xffff...`) inputs and zero value (`0x0000...`) inputs.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


recmo:
I'll fix CI errors and merge.

PR #233, opened 2023-04-19
fix: require 0x prefix for human readable bits

Rjected:
## Motivation

In reth we are using a `B256` as our `H256` type, in both RPC and internally. It seems that the `Bits` type, like the `Uint` type, will accept non-`0x`-prefixed strings. When subjecting reth to client test suites, we've run into a test case that expects us to fail parsing if there is no `0x` prefix: https://github.com/paradigmxyz/reth/issues/2306

## Solution

Create a new `PrefixedStrVisitor` that requires a `0x` prefix for strings, which errors if there is no `0x` prefix. Otherwise, it delegates to the existing `StrVisitor`.

Tests are added which check that there is an error for non `0x` prefixed strings.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/support/serde.rs
+++ b/src/support/serde.rs
@@ -36,9 +36,10 @@ impl<const BITS: usize, const LIMBS: usize> Serialize for Uint<BITS, LIMBS> {
     }
 }
 
-/// Deserialize human readable hex strings or byte arrays into hashes.
-/// Hex strings can be upper/lower/mixed case, have an optional `0x` prefix, and
-/// can be any length. They are interpreted big-endian.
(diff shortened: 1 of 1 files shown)


Rjected:
OK, nevermind on this being the cause for the above bug. Somehow I didn't notice that we used a different type for our `H256`, so we actually do not use this type for deserialization. But if ruint wants to require `0x` prefixes for bits, we can iterate on this, otherwise feel free to close!

prestwich:
This would also be a breaking change, and we'd prefer to keep deserialization permissive

PR #246, opened 2023-05-18
feat: make serde::Deserialize impl more permissive

DaniPopes:
## Motivation

Based on #243, the actual changes are in 9bda599

Uses the `str::FromStr` implementation to deserialize strings, and `TryFrom` to deserialize `u64` and `u128`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


src/string.rs
```diff
@@ -170,11 +170,12 @@ impl<const BITS: usize, const LIMBS: usize> FromStr for Uint<BITS, LIMBS> {
     type Err = ParseError;
 
     fn from_str(src: &str) -> Result<Self, Self::Err> {
-        if src.len() >= 2 {
-            match &src[..2] {
-                "0x" | "0X" => return Self::from_str_radix(&src[2..], 16),
-                "0o" | "0O" => return Self::from_str_radix(&src[2..], 8),
-                "0b" | "0B" => return Self::from_str_radix(&src[2..], 2),
+        if src.len() >= 2 && src.is_char_boundary(2) {
```
prestwich, line 173: `src.len() >= 2` is unnecessary, as `is_char_boundary` will return `false` if length is less than 2

src/support/serde.rs
```diff
@@ -106,20 +106,28 @@ impl<'de, const BITS: usize, const LIMBS: usize> Deserialize<'de> for Bits<BITS,
     }
 }
 
-/// Serde Visitor for human readable formats
-struct StrVisitor<const BITS: usize, const LIMBS: usize>;
+/// Serde Visitor for human readable formats.
+///
+/// Accepts either a primitive number, a decimal or a hexadecimal string.
+struct HRVisitor<const BITS: usize, const LIMBS: usize>;
```
prestwich, line 112: nit: rust naming convention is to only capitalize first letter in acronyms. `HrVisitor`

src/support/serde.rs
```diff
@@ -209,12 +184,26 @@ mod tests {
...
         });
     }
 
+    #[test]
+    fn test_human_readable_de() {
+        let jason = r#"[
+            1,
+            "0x1",
+            "0o1",
+            "0b1"
```
prestwich, line 199: test cases for prefix without body `"0x", "0o", "0b"`

prestwich (changes requested):

src/support/serde.rs
DaniPopes: those and the empty string "" return 0, is this expected?

prestwich: It is expected based on the behavior of `from_str_radix` and `from_base_be`, yes

Changelog.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Use Ethereum `Quantity` encoding for serde serialization when human-readable
```
prestwich, line 14: needs changelog update

src/support/serde.rs
```diff
@@ -61,54 +106,41 @@ impl<'de, const BITS: usize, const LIMBS: usize> Deserialize<'de> for Bits<BITS,
...
-                return Err(Error::invalid_value(Unexpected::Str(value), &self));
-            }
-            limbs[i] = limb;
+    fn visit_str<E: Error>(self, value: &str) -> Result<Self::Value, E> {
+        // Shortcut for common case
+        if value == ZERO_STR {
+            return Ok(Uint::<BITS, LIMBS>::ZERO);
         }
-        if BITS > 0 && limbs[LIMBS - 1] > Self::Value::MASK {
+        // `ZERO_STR` is the only valid serialization of `Uint<0, 0>`, so if we
```
prestwich, line 135: is this true with the more-permissive deser? do we want to use less-permissive deser for the special case?

src/support/serde.rs
```diff
@@ -202,32 +251,7 @@ mod tests {
             serialized.push('0');
             serialized.push('"');
             let deserialized = serde_json::from_str::<Uint<BITS, LIMBS>>(&serialized);
-            assert!(deserialized.is_err());
-        });
-    }
-
-    #[test]
-    fn test_serde_zero_invalid_size_error() {
```
prestwich, line 210: can you comment here why this test is no longer necessary?

DaniPopes: Size was checked against pre-calculated length in hex, which is not possible to do anymore because we accept other formats as well

src/support/serde.rs
DaniPopes: Yes it is true, in the two lines right below this

prestwich: what i mean is that do we still require a canonical serialization for 0-bit numbers? or do we permissively accept `0b0`, `0o0` and `0` for this type?

prestwich (approved):

PR #253, opened 2023-05-30
fix: `from_base_be` overflow

DaniPopes:
## Motivation

`LIMBS` can be 0, which panics at runtime on subtraction overflow

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


prestwich (approved):
Thanks for fixing my mistake 🖖

PR #260, opened 2023-07-07
Making std optional.

lvella:
This PR creates a new cargo feature `std`, on by default, that can be disabled if the user wants to target an environment with no support for `std` library.

## Motivation

Crate is now supported in a larger number of targets, and is a step into fixing this issue: https://github.com/bluealloy/revm/issues/493.

## Solution

This crate doesn't really use any std feature (e.g., threads, IO, file, network). It is a pure computation library, and theoretically it could be pure `#![no_std]` with no need for a feature to enable "std". But unfortunately I hit into two issues:
- The fact that `thiserror` is currently std-only (https://github.com/dtolnay/thiserror/pull/211), and the solution to make `thiserror` no_std depends on a unstable rust feature (https://github.com/rust-lang/rust/issues/103765).
- Some floating point functions are only available in `std` (like `log2`, `abs`, `pow`, etc), this is an issue of LLVM relying on the system libraries implementation for these functions. This also happens with `memcpy` and integer division functions, which LLVM assumes to exist, and is solved by including these functions in `compiler_builtins` library that can be used by no_std projects. For some reason, the corresponding floating point functions are not in `compiler_builtins` yet (https://github.com/rust-lang/rust/issues/50145). Thus, when `std` is disabled, this PR disables modules `log`, `pow` and `root`, along with `From<f64>` and `From<f32>` implementations, all which relies on the missing floating point functions.

## PR Checklist

- [?] Added Tests
- [✓] Added Documentation
- [✓] Updated the changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1687,7 +1687,7 @@ dependencies = [
  "serde_json",
  "smallvec",
  "sqlx-core",
- "thiserror",
+ "thiserror-core",
  "valuable",
  "zeroize",
 ]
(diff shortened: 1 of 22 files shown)


prestwich:
> The fact that thiserror is currently std-only (https://github.com/dtolnay/thiserror/pull/211), and the solution to make thiserror no_std depends on a unstable rust feature (https://github.com/rust-lang/rust/issues/103765).

Yes, you can't use `thiserror`, you have to implement `core::fmt::Display` manually, and then `std::error::Error` behind a cfg flag. It's a bit of a pain but not so bad.

Wizdave97:
> > The fact that thiserror is currently std-only ([dtolnay/thiserror#211](https://github.com/dtolnay/thiserror/pull/211)), and the solution to make thiserror no_std depends on a unstable rust feature ([rust-lang/rust#103765](https://github.com/rust-lang/rust/issues/103765)).
> 
> Yes, you can't use `thiserror`, you have to implement `core::fmt::Display` manually, and then `std::error::Error` behind a cfg flag. It's a bit of a pain but not so bad.

@lvella @prestwich  mind if I finish this up since we need it urgently on our end

gakonst:
Superseded in https://github.com/recmo/uint/pull/274, we'll get it done today. Thank you.

PR #264, opened 2023-07-17
feat: merge alloy-rs/ruint (ruint2), take 2

DaniPopes:
## Motivation

Internal changes:
- improve uint features and deps
  - remove build.rs feature "detection" by unifying cfgs into `feature = "nightly"`: https://github.com/recmo/uint/blob/35167fbf0406c8f925608225225f91f5612e371b/build.rs#L4C8-L4C76
- `git rm Cargo.lock`
- update all dependencies to latest versions
- remove profile overrides (#247)
- remove "dyn" stuff, since it's wholly incomplete
- bumps `ruint` to `1.9.0`, and `ruint-macro` to `1.1.0`

Dependabot:
- Closes #221
- Closes #251
- Closes #252
- Closes #258
- Closes #259

Other:
- Closes #247

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


Cargo.toml
```diff
@@ -1,80 +1,112 @@
...
-[package.metadata.docs.rs]
-rustdoc-args = ["--cfg", "has_cfg_doc", "--html-in-header", ".cargo/katex-header.html"]
-all-features = true
+[features]
+# nightly-only features
+nightly = []
+generic_const_exprs = ["nightly"]
+
+# support
+alloy-rlp = ["dep:alloy-rlp"]
```
prestwich, line 94: why are the support features features now explicit?

src/support/mod.rs
```diff
@@ -1,20 +1,23 @@
...
-mod rand;
-mod rlp;
-mod scale;
-mod serde;
-mod sqlx;
-mod valuable;
-mod zeroize;
+//! Support for external crates.
+
+pub mod alloy_rlp;
```
prestwich, line 3: do any of these modules contain exported types?

DaniPopes: Some of them, should I make the empty ones private?

Cargo.toml
DaniPopes: Better to be explicit, and helps in future with chaining on other features that it should enable (eg std, alloc, serde...)

src/support/mod.rs
prestwich: looks like only sqlx & scale have exported items. lets keep the other ones private to avoid cluttering user namespace

prestwich (approved):

PR #274, opened 2023-07-27
feat: no_std

DaniPopes:
## Motivation

Make `ruint` `no_std`-compatible.

Supersedes #260.

## Solution

See original description in #260.
Since `core::error::Error` is unstable, we instead implement the formatting using `core::fmt` and gate `impl std::error::Error` with `#[cfg(feature = "std")]`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,6 @@ serde_json = "1.0"
 [features]
 default = ["std"]
 std = [
-    "thiserror",
     "alloy-rlp?/std",
     "ark-ff-03?/std",
     "ark-ff-04?/std",
@@ -135,5 +134,5 @@ serde = ["dep:serde"]
 valuable = ["dep:valuable"]
 zeroize = ["dep:zeroize"]
 
-postgres = ["dep:postgres-types", "dep:bytes", "std"]
-sqlx = ["dep:sqlx-core", "std"]
+postgres = ["dep:postgres-types", "dep:bytes", "std", "dep:thiserror"]
+sqlx = ["dep:sqlx-core", "std", "dep:thiserror"]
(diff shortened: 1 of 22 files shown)


Cargo.toml
```diff
@@ -38,30 +38,34 @@ ruint-macro = { version = "1.1.0", path = "ruint-macro" }
 name = "bench_uint"
 harness = false
 path = "benches/bench.rs"
+required-features = ["std"]
 
 [dependencies]
 ruint-macro.workspace = true
 
-thiserror = "1.0"
+thiserror = { version = "1.0", optional = true }
```
prestwich, line 46: is this still used anywhere?

src/from.rs
```diff
@@ -32,51 +32,90 @@
...
+
+impl<T> fmt::Display for ToUintError<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::ValueTooLarge(bits, _) => write!(f, "Value is too large for Uint<{bits}>"),
+            Self::ValueNegative(bits, _) => {
+                write!(f, "Negative values can not be represented as Uint<{bits}>")
+            }
+            Self::NotANumber(bits) => {
+                write!(f, "'Not a number' (NaN) not be represented as Uint<{bits}>")
```
prestwich, line 65: nit: cannot

Cargo.toml
DaniPopes: postgres and sql

DaniPopes:
Coverage is actually unchanged because it didn't count code generated by the macro

prestwich (approved):
lgtm

PR #286, opened 2023-07-29
feat: check from base len to prevent consuming large iters

### LINKED ISSUE #279: OPT:`from_base_le` recursion depth check
as of #277 `from_base_le` uses recursion to stack allocate the contents of an iterator. However, there's no check on the recursion depth, allowing arbitrary-length iterators to be unpacked onto the stack. Extremely large iterators will overflow the stack this way. It also may unpack far more iterator items than the `Uint` can fit

- Compute  the maximum iterations using `BITS` and `base`
- Pass a recursion depth and increment at each recursing layer
- Once the recursion has reached a depth where the target BITS cannot fit the number of digits in `base`, return an error

prestwich:
Draft PR for discussion purposes:

## Motivation

closes #279

Currently both `from_base_be` and `from_base_le` accept more digits then can fit in the target `Uint`. This leads to extra iteration/recursion. If all extra digits are 0, there is unbounded extra iteration/recursion, which results in degenerate cases:

```
// infinite loop
Uint::<4, 1>::from_base_be(10, std::iter::repeat(0u64))

// overflows the stack due to recursion
Uint::<4, 1>::from_base_le(10, std::iter::repeat(0u64))
```

## Solution

Compute `max_digits` and bound the `from_base_*e` logic to operating on that many digits. Overflow if more digits are supplied.

However, this is a **breaking change**, as `from_base_be` is used in `from_str_radix` and `from_str`, and currently permits non-minimal string encodings. 

```
Uint::<16, 1>::from_str("0x00000000000001").unwrap();
```

**Open Question:**
Can we both
- allow non-minimal byte encodings (where all extra bytes are 0)
- avoid infinite iteration

**Possible solutions:**
- use `size_hint` on the iterators, and error if no upper bound established
- allow up to `N` extra 0s in addition to the meaningful digits. Error if more than `N`

These would still be breaking, but only in much more rare cases. And we would probably be comfortable doing that as a patch

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/base_convert.rs
+++ b/src/base_convert.rs
@@ -125,9 +125,24 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
             return Err(BaseConvertError::InvalidBase(base));
         }
 
-        let mut tail = digits.into_iter();
-        match tail.next() {
-            Some(digit) => Self::from_base_le_recurse(digit, base, &mut tail),
+        // we compute this by dividing the number of bits in the Uint by the
+        // number of bits in each word, rounding up.
+        let max_digits: usize = BITS / (usize::BITS - base.leading_zeros()) as usize + 1;
+
+        let mut iter = digits.into_iter();
+        let mut digits = iter.by_ref().take(max_digits);
+
+        match digits.next() {
+            Some(digit) => {
+                let res = Self::from_base_le_recurse::<I::IntoIter>(digit, base, digits)?;
+
+                // if the iter was not drained during the recursion process,
+                // then we have overflowed
+                if iter.next().is_some() {
+                    return Err(BaseConvertError::Overflow);
+                }
+                Ok(res)
+            }
             None => Ok(Self::ZERO),
         }
     }
@@ -141,7 +156,7 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
     fn from_base_le_recurse<I: Iterator<Item = u64>>(
         digit: u64,
         base: u64,
-        tail: &mut I,
+        mut tail: core::iter::Take<&mut I>,
     ) -> Result<Self, BaseConvertError> {
         if digit > base {
             return Err(BaseConvertError::InvalidDigit(digit, base));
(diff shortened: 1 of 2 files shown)


recmo:
> Currently both from_base_be and from_base_le accept more digits then can fit in the target Uint. This leads to extra iteration/recursion. If all extra digits are 0, there is unbounded extra iteration/recursion, which results in degenerate cases:

I'd consider this intentional. Leading zeros are supported by the base convertor and they make mathematical sense. If you feed it an infinite string of digits than I'd expect it to take infinite time. That's IMO not a bug but a consequence of the inversion of control created by Rust's iterators. If you give it an iterator that panics it will panic. If you give it an iterator that launches rockets, it will launch rockets.

We can look into optimizing the leading zero case. Handling them should be O(n) time, O(1) space.

prestwich:
It makes sense from a mathematical perspective. On the other hand, we can infer that it is never a sane user's intention to trigger this behavior. We also know that after the max digits is reached, any further non-zero digit is guaranteed to cause an overflow. It seems like we can/should use this knowledge to prevent or handle invalid use of the API

recmo:
Plenty of sane usecase where this would be useful, as it's pretty common to store numbers with an excessive amount of leading zeros as a way of padding. For example Solidity's ABI likes to store U160 padded up to U256. It would be nice if we can parse the U160 directly from the whole hex string and have it handle overflow correctly. The alternative would be an intermediate U256 and try_into, or messing with substrings and forgetting to check that the lead is indeed zero.

I'm proposing an alternate implementation of `from_base_le` that won't run out of memory/stack in #293 .

prestwich:
The sane usecases are bounded at (generously) 100 extra digits, and there's no sane usecase for infinitely looping. It's never good behavior for a library to hang indefinitely on valid input

Maybe this is a place to use `size_hint` and error out if no upper bound is established?

recmo:
Can you give me an example of a well-written rust library that puts similar bounds on user provided iterators? For example the following also hangs:

```rust
fn main() {
    println!("{:?}", core::iter::repeat(0_u64).min());
}
```

And here it doesn't even need to hang, because it could short-circuit as soon as it sees `u64::MIN`.

prestwich:
closing as there's no pressing need and I don't have a good plan anyway :)

PR #289, opened 2023-07-30
fix: support features

DaniPopes:
## Motivation

Some support features fail to compile when on "default_features = false" because they don't enable "alloc".

This wasn't caught by CI because running "--all-targets" implicitly enables "std" due to the self-dependency (ruint depends on ruint as a dev-dependency with extra features enabled).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


prestwich (changes requested):
needs cargo hack fixed

prestwich (approved):

PR #292, opened 2023-08-03
Feat: Implement `Pod` trait on Uint types

0xKitsune:
## Motivation

This PR implements the [`Pod` trait](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html) on `Uint` types where the size is a multiple of 64, up to 1024. This allows `Uint` types within this range to be used in instances where a `Pod` trait bound exists.

## Solution

This implementation uses a simple macro to implement the `Pod` trait on the following Uint sizes listed below.

```rust

macro_rules! impl_pod {
    ($(($bits:expr, $limbs:expr)),+ $(,)?) => {
        $(
            unsafe impl Pod for Uint<{$bits}, $limbs> {}
        )+
    };
}


impl_pod! {
    (64, 1),
    (128, 2),
    (192, 3),
    (256, 4),
    (320, 5),
    (384, 6),
    (448, 7),
    (512, 8),
    (576, 9),
    (640, 10),
    (704, 11),
    (768, 12),
    (832, 13),
    (896, 14),
    (960, 15),
    (1024, 16),
}
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,8 @@ named feature flag.
 * [`pyo3`](https://docs.rs/pyo3): Implements the [`ToPyObject`](https://docs.rs/pyo3/latest/pyo3/conversion/trait.ToPyObject.html), [`IntoPy`](https://docs.rs/pyo3/latest/pyo3/conversion/trait.IntoPy.html) and [`FromPyObject`](https://docs.rs/pyo3/latest/pyo3/conversion/trait.FromPyObject.html) traits.
 * [`parity-scale-codec`](https://docs.rs/parity-scale-codec): Implements the [`Encode`](https://docs.rs/parity-scale-codec/latest/parity_scale_codec/trait.Encode.html), [`Decode`](https://docs.rs/parity-scale-codec/latest/parity_scale_codec/trait.Decode.html), [`MaxEncodedLen`](https://github.com/paritytech/parity-scale-codec/blob/47d98a1c23dabc890fdb548d115a18070082c66e/src/max_encoded_len.rs) and [`HasCompact`](https://docs.rs/parity-scale-codec/latest/parity_scale_codec/trait.HasCompact.html) traits.
 * [`bn-rs`](https://docs.rs/bn-rs/latest/bn_rs/): Implements conversion to/from the [`BN`](https://docs.rs/bn-rs/latest/bn_rs/struct.BN.html) and [`BigNumber`](https://docs.rs/bn-rs/latest/bn_rs/struct.BigNumber.html).
+* [`bytemuck`](https://docs.rs/bytemuck): Implements the [`Pod`](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html) and [Zeroable](https://docs.rs/bytemuck/latest/bytemuck/trait.Zeroable.html) traits for [`Uint`] where the size is a multiple of 64, up to 1024. This allows `Uint` to be used where a `Pod` trait bound exists.
+
 
 ## Building and testing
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -151,6 +151,8 @@ named feature flag.
...
 * [`parity-scale-codec`](https://docs.rs/parity-scale-codec): Implements the [`Encode`](https://docs.rs/parity-scale-codec/latest/parity_scale_codec/trait.Encode.html), [`Decode`](https://docs.rs/parity-scale-codec/latest/parity_scale_codec/trait.Decode.html), [`MaxEncodedLen`](https://github.com/paritytech/parity-scale-codec/blob/47d98a1c23dabc890fdb548d115a18070082c66e/src/max_encoded_len.rs) and [`HasCompact`](https://docs.rs/parity-scale-codec/latest/parity_scale_codec/trait.HasCompact.html) traits.
 * [`bn-rs`](https://docs.rs/bn-rs/latest/bn_rs/): Implements conversion to/from the [`BN`](https://docs.rs/bn-rs/latest/bn_rs/struct.BN.html) and [`BigNumber`](https://docs.rs/bn-rs/latest/bn_rs/struct.BigNumber.html).
+* [`bytemuck`](https://docs.rs/bytemuck): Implements the [`Pod`](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html) and [Zeroable](https://docs.rs/bytemuck/latest/bytemuck/trait.Zeroable.html) traits for [`Uint`] where the size is a multiple of 64, up to 1024. This allows `Uint` to be used where a `Pod` trait bound exists.
```
recmo, line 154: `Zeroable` should be in quotes.

src/support/bytemuck.rs
```diff
@@ -0,0 +1,66 @@
...
+#![cfg_attr(docsrs, doc(cfg(feature = "bytemuck")))]
+
+use bytemuck::Pod;
+use ruint::Uint;
+
+// Implement the `Pod` trait for `Uint` types with a size that is a multiple of 64, up to 1024.
+macro_rules! impl_pod {
+    ($($bits:expr),+ $(,)?) => {
+        $(
+            unsafe impl<const LIMBS: usize> bytemuck::Zeroable for Uint<{$bits}, LIMBS> {}
```
recmo, line 12: `Zeroable` should be defined once generically for all Uint. It does not require the BITS to be a multiple of `64` so should not be part of this macro.

src/support/bytemuck.rs
```diff
@@ -0,0 +1,66 @@
...
+            unsafe impl<const LIMBS: usize> Pod for Uint<{$bits}, LIMBS> where
+                [(); {$bits % 64}]: ,
```
recmo, line 14: What does this clause do?

src/support/bytemuck.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use bytemuck::Pod;
+use ruint::Uint;
+
+// Implement the `Pod` trait for `Uint` types with a size that is a multiple of 64, up to 1024.
+macro_rules! impl_pod {
+    ($($bits:expr),+ $(,)?) => {
+        $(
+            unsafe impl<const LIMBS: usize> bytemuck::Zeroable for Uint<{$bits}, LIMBS> {}
+            unsafe impl<const LIMBS: usize> Pod for Uint<{$bits}, LIMBS> where
```
recmo, line 13: In other places (like aliases.rs) we let the macro run over (BITS, LIMBS) pairs. I think that's cleaner than making one argument generic.

recmo (changes requested):

src/support/bytemuck.rs
0xKitsune: Originally, I was trying to find a way to enforce a compile time check to ensure that `bits` was a multiple of 64, however this is not achieving that and I have simplified to the following:

```rust
macro_rules! impl_pod {
    ($(($bits:expr, $limbs:expr)),+ $(,)?) => {
        $(
            unsafe impl Pod for Uint<{$bits}, $limbs> {}
        )+
    };
}
```

src/support/bytemuck.rs
```diff
@@ -0,0 +1,79 @@
...
+// Implement the `Pod` trait for `Uint` types with a size that is a multiple of 64, up to 1024.
+macro_rules! impl_pod {
+    ($(($bits:expr, $limbs:expr)),+ $(,)?) => {
+        $(
+            unsafe impl Pod for Uint<{$bits}, $limbs> {}
+        )+
+    };
+}
+
+impl_pod! {
```
prestwich, line 20: is it possible to replace this with a `generic_const_expr` usage when `cfg(feature = "nightly")` ?

src/support/bytemuck.rs
```diff
@@ -0,0 +1,82 @@
...
+
+// Implement Zeroable for all `Uint` types.
+unsafe impl<const BITS: usize, const LIMBS: usize> Zeroable for Uint<{ BITS }, { LIMBS }> {}
+
+// Implement the `Pod` trait for `Uint` types with a size that is a multiple of
+// 64, up to 1024. Note that implementors must have a size that is divisible by
+// 64, and using `Uint` sizes not divisible by 64 would violate Pod's
+// guarantees potentially leading to undefined behavior.
+macro_rules! impl_pod {
+    ($(($bits:expr, $limbs:expr)),+ $(,)?) => {
```
prestwich, line 16: nit: these are really `literal`, as non-literal `expr` are invalid in this context. no change needed

src/support/bytemuck.rs
prestwich: i don't consider this a blocker

recmo (approved):

PR #293, opened 2023-08-03
Constant space from base conversions

recmo:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/algorithms/mul.rs
+++ b/src/algorithms/mul.rs
@@ -224,7 +224,18 @@ fn mac(lhs: &mut u64, a: u64, b: u64, c: u64) -> u64 {
     prod.high()
 }
 
-/// Computes `lhs += a * b` and returns the borrow.
+/// Computes `lhs *= a` and returns the borrow.
+pub fn mul_nx1(lhs: &mut [u64], a: u64) -> u64 {
+    let mut carry = 0;
+    for lhs in lhs.iter_mut() {
+        let product = u128::muladd(*lhs, a, carry);
+        *lhs = product.low();
+        carry = product.high();
+    }
+    carry
+}
+
+/// Computes `lhs += a * b` and returns the carry.
 ///
 /// Requires `lhs.len() == a.len()`.
 ///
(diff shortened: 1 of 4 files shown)


src/base_convert.rs
```diff
@@ -124,36 +101,42 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
...
+        let mut iter = digits.into_iter();
+        let mut result = Self::ZERO;
+        let mut power = Self::from(1);
+        while let Some(digit) = iter.next() {
+            if digit >= base {
+                return Err(BaseConvertError::InvalidDigit(digit, base));
+            }
+            let term = power.checked_mul(Self::from(digit)).ok_or(BaseConvertError::Overflow)?;
+            result = result.checked_add(term).ok_or(BaseConvertError::Overflow)?;
+            if let Some(next_power) = power.checked_mul(Self::from(base)) {
```
recmo, line 125: These three lines can be combined into a simultaneous iteration over both `power` and `result`. It will similar to, but a bit more complex than the loop in `from_base_be`. As it is currently it doesn't make good use of multiplicants being single limbs.

prestwich:
approach is good, i think we still disagree on the behavior wrt infinite iterators

recmo:
This is mostly orthogonal to how long iterators are handled and should make things strictly better. Ready for review.

src/algorithms/mul.rs
```diff
@@ -224,7 +224,18 @@ fn mac(lhs: &mut u64, a: u64, b: u64, c: u64) -> u64 {
     prod.high()
 }
 
-/// Computes `lhs += a * b` and returns the borrow.
+/// Computes `lhs *= a` and returns the carry.
+pub fn mul_nx1(lhs: &mut [u64], a: u64) -> u64 {
+    let mut carry = 0;
+    for lhs in lhs.iter_mut() {
```
prestwich, line 230: if `lhs` is empty this will return 0. is this intended, or should it return `a`?

recmo: > my belief is that a 0-byte uint always represents the number 0

This is correct. See for example `U0::ZERO`. (This is also the reason why no `Uint::ONE` can exits)

prestwich (approved):

PR #303, opened 2023-09-03
Update aliases.rs

Qumeric:
Add U784 and B784 aliases

This is useful because 96 bytes is the length `BlsSignature` used in, for example, MEV-Boost (relays accept signatures in this format).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `bytemuck` feature ([#292])
+- `Uint::is_zero() -> bool` ([#296])
 - `num-traits` features ([#298])
 
 [#292]: https://github.com/recmo/uint/pulls/292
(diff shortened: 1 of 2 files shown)


src/aliases.rs
```diff
@@ -56,6 +56,7 @@ alias! {
     U384 B384 (384, 6);
     U448 B448 (448, 7);
     U512 B512 (512, 8);
+    U784 B784 (784, 12);
```
prestwich, line 59: the `LIMBS` param needs to be 13. `64 * 12 == 768`, which will not quite fit 784 bits

Qumeric: yeah it should be U768 and B768, sorry

prestwich: ah cool. mind pushing a commit to change that?

prestwich:
succeeded here: https://github.com/recmo/uint/pull/310

because i messed up the branch updating :D

Qumeric:
Thanks @prestwich , sorry for being a bit sloppy here

prestwich:
Nah i was also a bit sloppy lol. We're dealing with some dep issues right now, so this will go in once those are resolved

PR #326, opened 2023-10-01
chore: loops clean up

DaniPopes:
## Motivation

- Use normal range loop where it makes sense.
- Use same impl as `Ord for [T]` for `algorithms::cmp`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -48,7 +48,8 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
     #[cfg(feature = "alloc")]
     #[must_use]
     #[inline]
-    pub const fn as_le_bytes(&self) -> Cow<'_, [u8]> {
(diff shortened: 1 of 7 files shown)


src/bytes.rs
```diff
@@ -57,8 +57,10 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         #[cfg(target_endian = "big")]
         return Cow::Owned({
             let mut cpy = *self;
-            cpy.limbs.iter_mut().for_each(|limb| limb.reverse_bits());
-            slice::from_raw_parts(cpy.limbs.as_ptr().cast(), Self::BYTES).to_vec()
```
DaniPopes, line 61: This didn't compile, also can't make the function const.

prestwich (approved):

PR #329, opened 2023-10-13
feat: make `{from,to}_{b,l}e_bytes` const

DaniPopes:
## Motivation

Const impl + some perf gains

## Solution

Hand roll use of iterators, for loops and `slice::reverse`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Uint::is_zero() -> bool` ([#296])
 - `num-traits` features ([#298])
 - `U768` alias ([#310])
-- Improve `add` and `sub` performance ([#316])
-- Make `add` and `sub` functions `const` ([#324])
-- Make `{from,to}_{b,l}e_bytes` const ([#329])
+- Improved `add` and `sub` performance ([#316])
+- Made `add` and `sub` functions `const` ([#324])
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,17 +11,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `U768` alias ([#310])
 - `bytemuck` feature ([#292])
 - `Uint::is_zero() -> bool` ([#296])
 - `num-traits` features ([#298])
+- `U768` alias ([#310])
```
prestwich, line 17: looks like i forgot to add the link to the block below. mind doing that as a drive-by?

src/bytes.rs
```diff
@@ -175,57 +192,85 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         bytes
     }
 
-    /// Creates a new integer from a little endian stream of bytes.
+    /// Converts a big-endian byte array of size exactly
+    /// [`Self::BYTES`] to [`Uint`].
```
prestwich, line 196: any reason for reordering methods?

DaniPopes: consistency

prestwich: ya, just makes the functional review harder

prestwich (approved):
Very nice

PR #335, opened 2023-10-19
fix: disallow leading zeros in RLP decoding

Rjected:
## Motivation

Currently `Uint` RLP decoding ignores leading zeros.

## Solution

Disallow leading zeros.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `leading_ones` failed for non-aligned sizes.
+- Restricted RLP decoding to match the RLP spec and disallow leading zeros ([#335])
+
+[#335]: https://github.com/recmo/uint/pulls/335
 
 ## [1.10.1] - 2023-07-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Restricted RLP decoding to match the RLP spec and disallow leading zeros ([#335])
+
+[#335]: https://github.com/recmo/uint/pulls/335
+
 ## [1.11.0] - 2023-10-27
 
 ### Added
@@ -32,9 +38,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `leading_ones` failed for non-aligned sizes.
-- Restricted RLP decoding to match the RLP spec and disallow leading zeros ([#335])
-
-[#335]: https://github.com/recmo/uint/pulls/335
 
 ## [1.10.1] - 2023-07-30
 
--- a/src/support/alloy_rlp.rs
+++ b/src/support/alloy_rlp.rs
@@ -13,7 +13,7 @@ const MAX_BITS: usize = 55 * 8;
 
 /// Allows a [`Uint`] to be serialized as RLP.
 ///
-/// See <https://eth.wiki/en/fundamentals/rlp>
+/// See <https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/>
 impl<const BITS: usize, const LIMBS: usize> Encodable for Uint<BITS, LIMBS> {
     #[inline]
     fn length(&self) -> usize {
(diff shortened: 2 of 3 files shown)


prestwich:
This would be a breaking behavior change. Is there a motivating reason?

Rjected:
> This would be a breaking behavior change. Is there a motivating reason?

Sorry, should have been more specific in the PR description. The main motivation is that leading zeros on RLP uints is defined to be invalid, and encoding would also never output uints encoded this way. This also makes writing certain kinds of fuzz tests possible, specifically fuzz tests that:
* Take fuzz data as input
* Decode TX or other struct with an int
* if successful, encode struct
* compare encoding output to bytes input

If we accept invalid RLP input, and are more permissive in decoding than encoding as a result, these fuzz tests quickly fail and do not act as meaningful fuzz tests.

Rjected:
@prestwich just bumping, any thoughts on this given the context?

prestwich:
The ethereum.org spec is clear as mud, but the yellowpaper appears to be unambiguous about this ([see (197)](https://ethereum.github.io/yellowpaper/paper.pdf#page=20)) . Accepting the non-minimal uint is off-spec, so we should fix it. However, because this silently changes behavior in a way that breaks downstream users, we should release it as a major version bump

prestwich (changes requested):
Update the changelog :)

src/support/alloy_rlp.rs
recmo, line 1: Link to spec on line 16 and 75 is outdated. Should be <https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/>. Please fix.

src/support/alloy_rlp.rs
```diff
@@ -77,6 +77,13 @@ impl<const BITS: usize, const LIMBS: usize> Decodable for Uint<BITS, LIMBS> {
     #[inline]
     fn decode(buf: &mut &[u8]) -> Result<Self, Error> {
         let bytes = Header::decode_bytes(buf, false)?;
+
+        // We only need to check if the first byte is zero to make sure there are no
+        // leading zeros
```
recmo, line 82: Comment should mention that this is according to the RLP spec.

src/support/alloy_rlp.rs
```diff
@@ -77,6 +77,13 @@ impl<const BITS: usize, const LIMBS: usize> Decodable for Uint<BITS, LIMBS> {
...
+        if !bytes.is_empty() && bytes[0] == 0 {
+            return Err(Error::LeadingZero);
+        }
```
recmo, line 85: RLP spec explicitly says

> integers must be represented in big-endian binary form with no leading zeroes (thus making the integer value zero equivalent to the empty byte array).

meaning that the old accepting behaviour was not according to spec and a bug. Fixing bugs is not a breaking change.

There's 'robustness principle' argument that could be made to make this error optionally a warning (i.e. introduce a non-strict mode). But that seems more effort than it's worth. Unless someone volunteers the strict mode should be default.

recmo (approved):
LGTM. Bug fix. Not a breaking change.

recmo:
> If we accept invalid RLP input, and are more permissive in decoding than encoding as a result, these fuzz tests quickly fail and do not act as meaningful fuzz tests.

This proposed test assumes that encoding is canonical. This assumption is only true when leading zeros are rejected.

> [ethereum.org] and [yellow paper]

Both yellow paper and ethereum.org spec clearly state leading zeros are an error. While neither mention it explicitly, the implicit goal is to make it such that there is only one canonical RLP encoding. This is important since it RLP encoding is often used to hash data structures. Having multiple valid hashes for the same data structure leads to all forms of madness.

recmo:
> However, because this silently changes behavior in a way that breaks downstream users, we should release it as a major version bump

The accepting behavior was a bug, any downstream users depending on this bug are in error (as is clear by the RLP spec). IMO this a bug fix and not a breaking change.

prestwich:
I've also submitted a PR to the RLP spec doc on the Ethereum website to disambiguate this

the worst part of this is how it breaks the layer boundary between RLP and "higher-order" protocols. You cannot accurately decode RLP without knowing whether the higher-order protocol contains positive integers

https://github.com/ethereum/ethereum-org-website/pull/11532

src/support/alloy_rlp.rs
Rjected: fixed

src/support/alloy_rlp.rs
Rjected: quoted directly

DaniPopes (changes requested):
Can you also update fastrlp? It should be an easy copy-paste

Rjected:
> Can you also update fastrlp? It should be an easy copy-paste

good catch, will do

prestwich (changes requested):
one more changelog nit

CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `leading_ones` failed for non-aligned sizes.
+- Restricted RLP decoding to match the RLP spec and disallow leading zeros ([#335])
```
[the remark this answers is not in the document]
Rjected: fixed!

prestwich (approved):
🙌

PR #336, opened 2023-10-20
feat: implement `TryFrom<f64>` for no_std

DaniPopes:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


DaniPopes:
Actually this is probably wrong, might be worth adding [`libm`](https://docs.rs/libm/latest/libm/) for supporting missing float functions

PR #341, opened 2023-10-31
release: roll rlp fix in 1.11.0

gakonst:
as title -- will tag main with 1.11.0 and then release on crates


prestwich (commented):
LGTM

prestwich (changes requested):
wait, there are now 2 `# Fixed`. pls dedup

prestwich (approved):

PR #344, opened 2023-11-11
feat: ssz encode and decode for uint

loocapro:
## Motivation
Completing SSZ implementation also for Uint types: https://github.com/alloy-rs/core/pull/407

## Solution
Implementing Encode and Decode traits from https://github.com/sigp/ethereum_ssz

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,9 +108,9 @@ std = [
     "rlp?/std",
     "serde?/std",
     "valuable?/std",
-    "zeroize?/std",
-    "ethereum_ssz",
+    "zeroize?/std"
 ]
+ssz = ["std", "dep:ethereum_ssz"]
 alloc = ["proptest?/alloc", "rand?/alloc", "serde?/alloc", "valuable?/alloc", "zeroize?/alloc"]
 
 # nightly-only features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ serde = { version = "1", optional = true, default-features = false }
 valuable = { version = "0.1", optional = true, default-features = false }
 zeroize = { version = "1.6", optional = true, default-features = false }
 bytemuck = { version = "1.13.1", optional = true, default-features = false }
-ethereum_ssz = { version = "0.5.3" }
+ethereum_ssz = { version = "0.5.3", optional = true, default-features = false }
 
 # postgres
 bytes = { version = "1.4", optional = true }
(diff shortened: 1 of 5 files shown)


src/support/ssz.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
+        if bytes.len() > nbytes(BITS) {
+            return Err(DecodeError::InvalidByteLength {
+                len:      bytes.len(),
+                expected: nbytes(BITS),
+            });
+        }
+        let mut buf = [0u8; nbytes(BITS)]; // generic parameters may not be used in const operations
```
loocapro, line 41: Having some issue on using a generic parameter as a constant here.

Any advice @prestwich or @DaniPopes ?

prestwich: You should be able to do `Ok(Self::from_le_slice(bytes))`, no?

prestwich (changes requested):

Cargo.toml
```diff
@@ -108,6 +109,7 @@ std = [
     "serde?/std",
     "valuable?/std",
     "zeroize?/std",
+    "ethereum_ssz",
```
DaniPopes, line 112: This is not right

DaniPopes (changes requested):
this needs a separate explicit feature instead of the implicit one

src/support/ssz.rs
```diff
@@ -32,15 +32,7 @@ impl<const BITS: usize, const LIMBS: usize> Decode for Uint<BITS, LIMBS> {
...
-        if bytes.len() > nbytes(BITS) {
-            return Err(DecodeError::InvalidByteLength {
-                len:      bytes.len(),
-                expected: nbytes(BITS),
-            });
-        }
-        let mut buf = [0u8; nbytes(BITS)]; // generic parameters may not be used in const operations
-        buf[..bytes.len()].copy_from_slice(bytes);
-        Ok(Self::from_le_bytes(buf))
+        Ok(Self::from_le_slice(bytes))
```
prestwich, line 35: still want the length check with custom error, as `from_le_slice` will panic on overflows
```
        if bytes.len() > nbytes(BITS) {
            return Err(DecodeError::InvalidByteLength {
                len:      bytes.len(),
                expected: nbytes(BITS),
            });
        }```

prestwich:
Looks like tiny-keccak is `CC0 1.0` which is odd. We'll need to explicitly add that license to the approved list in `deny.toml`

prestwich (approved):

prestwich:
okay last thing I promise, please update CHANGELOG.md 🖖

PR #347, opened 2024-01-03
fix: modify `overflowing_shr` to fix overflow check

### LINKED ISSUE #337: `overflowing_shr` result is incorrect
**Version**
1.10.1

**Description**

https://github.com/recmo/uint/blob/666cd312be3810a1301257b1aaa40f03193788c7/src/bits.rs#L376-L381

The overflow result of `overflowing_shr` is incorrect.

The above can be modified as follows:
```rust
// Check for overflow
let mut overflow = false;
for i in 0..limbs {
overflow |= self.limbs[i] != 0;
}
overflow |= self.limbs[limbs] << (64 - bits) != 0;
```

-----

I tried this code:

```rust
let mut test_slice = [40u64];
let a = Uint::<64, 1>::from_limbs(test_slice);
println!("{:?}", a.overflowing_shr(1));

test_slice[0] = 41u64;
let b = Uint::<64, 1>::from_limbs(test_slice);
println!("{:?}", b.overflowing_shr(1));
```

I expected to see this happen: 
```
(0x0000000000000014_U64, false)
(0x0000000000000014_U64, true)
```

Instead, this happened: 
```
(0x0000000000000014_U64, true)
(0x0000000000000014_U64, true)
```

tcoratger:
## Motivation

`overflowing_shr` algorithm seems to give incorrect overflow results: #337.

Should close #337.

## Solution

- Added code to evaluate overflow for shift operation.
- This addition enables the identification of a floating-point outcome by isolating the most significant bit of the shifted limb.
- If the least significant bit of the shifted limb is non-zero, `overflow` is set to `true`, indicating a floating-point result.
- Enhances the functionality of the `overflowing_shr` function to differentiate between floating-point and integer outcomes after the shift operation.

This enhancement improves the functionality of the `overflowing_shr` method, providing a mechanism to determine whether the result represents a floating-point or integer value.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [X] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Overflow check in `overflowing_shr` implementation ([#347])
+
+[#347]: https://github.com/recmo/uint/pulls/347
+
 ## [1.11.1] - 2023-11-18
 
 ### Fixed
@@ -16,10 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Typo in `Shr` implementation ([#343])
 
 [#343]: https://github.com/recmo/uint/pulls/343
-
-- Overflow check in `overflowing_shr` implementation
-
-[#347]: https://github.com/recmo/uint/pulls/347
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


prestwich (approved):
Edited the changelog to move the line to the unreleased section.

prestwich:
this change has been published in v1.12.0 thanks!

PR #363, opened 2024-03-28
fix num-traits with no-std

xcshuan:
## Motivation

I ran into the problem below when I wanted to use num-traits in a no-std (riscv64imac-unknown-none-elf) environment.

```
Compiling ruint v1.12.1
error[E0412]: cannot find type `Vec` in this scope
  --> /home/xcshuan/.cargo/registry/src/github.com-1ecc6299db9ec823/ruint-1.12.1/src/support/num_traits.rs:80:18
   |
80 |     type Bytes = Vec<u8>;
   |                  ^^^ not found in this scope
   |
help: consider importing this struct
   |
7  + use alloc::vec::Vec;
   |

For more information about this error, try `rustc --explain E0412`.
error: could not compile `ruint` (lib) due to 1 previous error
```

## Solution

Add the following import code to `num_traits.rs`.

```
#[cfg(feature = "alloc")]
#[allow(unused_imports)]
use alloc::vec::Vec;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- add `alloc` requirement to `num-traits` feature  [#363]
+
+[#363]: https://github.com/recmo/uint/pull/363
+
 ## [1.12.1] - 2024-03-12
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):
Thanks! Can you also update the changelog? cc @prestwich

prestwich:
made a slight adjustment here:
- `num-traits` feature now enables `alloc`
- cfg on `use` statement changed to `not(feature = "std")` to prevent redundant import lints when standard is active

prestwich (approved):

PR #384, opened 2024-06-15
chore: allow Unicode-3.0 in deny.toml

DaniPopes:


prestwich (approved):
https://opensource.org/license/unicode-license-v3

PR #385, opened 2024-06-15
ci: temporarily disable codecov workflow

DaniPopes:


prestwich:
can you write the motivation here for future reader?

prestwich (approved):

DaniPopes:
It has been failing for a while due to (reason unknown)

PR #400, opened 2024-10-15
Bump sqlx 0.7 -> 0.8.2

Dzejkop:
## Motivation

Older versions of sqlx contain a [vulnerability](https://rustsec.org/advisories/RUSTSEC-2024-0363.html)

## Solution

This vulnerability has been patched in >=0.8.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for sqlx @ 0.8 ([#400])
+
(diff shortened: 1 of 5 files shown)


prestwich:
Thank you! In the past we've avoided breaking changes in feature subdeps as it would technically be a breaking API change for this crate. See the handling of ark-ff in `Cargo.toml` and `src/support/` Given that the previous version contains a vulnerability, we would consider a policy exception here

Would you mind fixing the fmt and clippy issues and then i'll review?

prestwich:
I've updated the changelog and set this to merge when checks pass, thank you!

Dzejkop:
Thanks!

PR #402, opened 2024-11-04
Optimize Montgomery multiplication

### LINKED ISSUE #284: `modular`: Make mul_redc alloc-free
*On 2023-07-29 @prestwich wrote in [`aaa37d2`](https://github.com/recmo/uint/commit/aaa37d2268b1a8996499953c84cf4f23474b1a67) “Merge pull request #277 from DaniPopes/no_alloc”:*

Make mul_redc alloc-free

```rust
    ///
    /// # Panics
    ///
    /// Panics if `inv` is not correct.
    #[must_use]
    #[cfg(feature = "alloc")] // TODO: Make mul_redc alloc-free
    pub fn mul_redc(self, other: Self, modulus: Self, inv: u64) -> Self {
        if BITS == 0 {
            return Self::ZERO;
        }
        assert_eq!(inv.wrapping_mul(modulus.limbs[0]), u64::MAX);

```
*From [`src/modular.rs:147`](https://github.com/recmo/uint/blob/aaa37d2268b1a8996499953c84cf4f23474b1a67/src/modular.rs#L147)*

recmo:
## Motivation

- Improves performance for `mul_redc`, which is critical to efficient prime field implementations.
- Fixes #284.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


recmo:
@prestwich Would love to see this merged!

src/algorithms/mul_redc.rs
```diff
@@ -1,70 +1,289 @@
...
+const fn carrying_add(lhs: u64, rhs: u64, carry: bool) -> (u64, bool) {
+    let (result, carry_1) = lhs.overflowing_add(rhs);
+    let (result, carry_2) = result.overflowing_add(carry as u64);
+    (result, carry_1 | carry_2)
+}
+
+// Helper while [Rust#85532](https://github.com/rust-lang/rust/issues/85532) stabilizes.
+#[inline]
+#[must_use]
+const fn borrowing_sub(lhs: u64, rhs: u64, borrow: bool) -> (u64, bool) {
```
DaniPopes, line 209: We define these exact functions in add.rs https://github.com/recmo/uint/blob/b91e6f710e42b893e695cae43ffab3d65740d881/src/add.rs#L61
We should probably move the definition to `algorithms` maybe?

src/algorithms/mul_redc.rs
```diff
@@ -1,70 +1,289 @@
...
-            carry >>= 64; // Sign extending shift
+
+    // Compute reduced product.
+    debug_assert!(carry_outer <= 1);
+    reduce1_carry(result, modulus, carry_outer > 0)
+}
+
+#[inline]
+#[must_use]
+fn less_than<const N: usize>(lhs: [u64; N], rhs: [u64; N]) -> bool {
```
DaniPopes, line 124: Can use `algorithms::cmp` https://github.com/recmo/uint/blob/b91e6f710e42b893e695cae43ffab3d65740d881/src/algorithms/mod.rs#L94

prestwich: also, what benefit would we get from monomorphizing here? would it be more/less efficient to do `for i in 0..N` instead of `zip`? I need to develop better intuition for this 🤔

recmo: I've found concrete performance to be better with `<const N: usize>(x: [u64; N])` rather than `(x: &[u64])`. In particular for Montgomery multiplication we benefit a lot from the compiler unrolling the loop when `N` is small and inlining the `modulus` when it is a compile time constant (which also eliminates the modulus-dependent branches).

src/algorithms/mul_redc.rs
recmo: I think my idea was to keep the `mul_redc` file self-contained (I've been copying it between projects), but there's no reason to do this here. I'll update.

src/algorithms/mul_redc.rs
recmo: I've found concrete performance to be better with `<const N: usize>(x: [u64; N])` rather than `(x: &[u64])`. In particular for Montgomery multiplication we benefit a lot from the compiler unrolling the loop when `N` is small and inlining the `modulus` when it is a compile time constant (which also eliminates the modulus-dependent branches).

recmo: In this case `less_than` was no longer used in production code, so no reason to keep it.

recmo:
I've found concrete performance to be better with `<const N: usize>(x: [u64; N])` rather than `(x: &[u64])`. In particular for Montgomery multiplication we benefit a lot from the compiler unrolling the loop when `N` is small and inlining the `modulus` when it is a compile time constant (which also eliminates the modulus-dependent branches).

I'm not sure how far we should push this observation through. E.g. would we want to do this for the division and GCD algorithm as well?

Right now the library is mostly instantiate with a small compile time `N`, typically `N=4`. But I'm also toying with the idea of adding support for proper dynamically sized uints at some point, in which case slice based algorithms are the only way.

prestwich:
CI failure is due to a backwards incompatible change in proptest, and can be smoothed over by pinning to 1.5.0. Gonna investigate what's wrong otherwise

prestwich:
> Right now the library is mostly instantiate with a small compile time N, typically N=4. For these cases the specialization would probably benefit performance. But I'm also toying with the idea of adding support for proper dynamically sized uints at some point, in which case slice based algorithms are the only way, and keeping both around violates DRY.

The idea being to allow expansion and do algorithm-switching based on the current size like go's `big.Int`?

> I've found concrete performance to be better with <const N: usize>(x: [u64; N]) rather than (x: &[u64]). In particular for Montgomery multiplication we benefit a lot from the compiler unrolling the loop when N is small and inlining the modulus when it is a compile time constant (which also eliminates the modulus-dependent branches).

my naive question is whether `for i in 0..N { match a[i] < b[i] { ... } }` will be optimized more readily than `for (left, right) in zip(a,b) { match left < right }`. I may sit down and investigate this at some point

prestwich:
> CI failure is due to a backwards incompatible change in proptest, and can be smoothed over by pinning to 1.5.0. Gonna investigate what's wrong upstream

i went ahead and pushed a commit to do this and opened #409, as I'm looking to merge this branch today

recmo:
> The idea being to allow expansion and do algorithm-switching based on the current size like go's big.Int?

There are two directions, `big.Int` is just one of them:

1. Compile time fixed, stack allocated: `[u64; N]` (current)
2. Compile time fixed, heap allocated: `Box<[u64; N]>`
3. Runtime fixed, heap allocated: `Box<[u64]>`
4. Runtime dynamic, heap allocated: `Vec<u64>`

When working with large-ish sizes, e.g. U4096, stack size may be limited and an approach 2-4 is required.

When implementing runtime negotiated cryptographic protocols (I stumbled on this in the `icao-9303` lib), a runtime size 3 or 4 is required.

To approximate natural numbers and have virtual unlimited size, 3 or 4 is required.

This is orthogonal to the problem of algorithm selection, which depends purely on size and would be the same for all approaches. Though practically, limited stack size prevents you from using approach 1 with sizes where anything other than base-case algorithms are relevant (GMP doesn't do fancy multiplication until 20-30 limbs on modern hw: https://gmplib.org/devel/thres/MUL_TOOM22_THRESHOLD)

prestwich (approved):

PR #404, opened 2024-11-08
feat: support for diesel @ 2.2

sinasab:
## Motivation

I noticed there were some comment notes in the repo indicating a desire for diesel support. I had implemented something similar using a newtype pattern in my own project, so I figured I'd attempt to contribute it back upstream.

## Solution

This PR implements the required traits for the generic uint types to work out of the box with diesel. The implementation closely matches the sqlx feature implementation in this repo. I have an example repo illustrating simple usage (schema, querying, inserting) @ https://github.com/sinasab/diesel_uint_test/.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

PR #403, opened 2024-11-04
feat: add support for fastrlp@0.4

Evalir:
## Motivation

Supersedes https://github.com/recmo/uint/pull/398, #401 

## Solution

Mirror ark-ff implementation of traits for multiple versions

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog

PR #406, opened 2024-11-27
Update deps

kpp:
Update deps.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,17 +65,17 @@ parity-scale-codec = { version = "3", optional = true, features = [
     "derive",
     "max-encoded-len",
 ], default-features = false }
-primitive-types = { version = "0.12", optional = true, default-features = false }
+primitive-types = { version = "0.13", optional = true, default-features = false }
 proptest = { version = "1.3", optional = true, default-features = false }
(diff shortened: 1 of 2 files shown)


prestwich:
wrt optional deps, our semver policy treats breaking changes as semver major. We typically update by adding support for specific versions (see #403 and the discussion in #404).  If you have a need for support for these crates at specific versions we can add those versions to the support directory, but modifying (removing) the existing support is a no go from our side

kpp:
Can you help me with that? I wanted to update deps to reduce the number of outdated deps in my project.

If I want to update `primitive-types`, `rlp` and `ethereum_ssz`, I have to add new features for those updated deps?

prestwich:
> Can you help me with that? I wanted to update deps to reduce the number of outdated deps in my project.

I think that #403 is a pretty good roadmap for achieving that. The basic steps are
- add the new dependency with version number in name (renaming the old one to include a version number too)
- add the new file to `support/`
- edit the existing support to use the renamed version of the old package
- do NOT edit the feature name

prestwich:
in the meantime, i'm closing this as `wontfix` as this PR would violate our semver policies. would love to see followup PRs tho :)

PR #416, opened 2024-12-17
feat: support for borsh @ 1.5

rtrombone:
## Motivation

Avoid having to implement a new type to handle `BorshSerialize` and `BorshDeserialize` for ruint types.

## Solution

This PR implements `BorshSerialize` and `BorshDeserialize` for `Uint` and `Bits` with basic tests.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog


prestwich (approved):
This seems reasonable to me. I am not an expert in borsh, so while I've tried to review the logic I am not 100% sure it's correct. Can you leave a 1-2 paragraph comment here about relevant particulars of borsh serialization in case I need to maintain this code in the future :)

rtrombone:
> This seems reasonable to me. I am not an expert in borsh, so while I've tried to review the logic I am not 100% sure it's correct. Can you leave a 1-2 paragraph comment here about relevant particulars of borsh serialization in case I need to maintain this code in the future :)

Of course! Borsh serializes integers in little-endian format. For `Uint` deserialization, I think iterating forward over the limbs and serializing each limb as little-endian (via `BorshDeserialize::deserialize_reader`) does the trick. Do you mind double-checking me?

For example, the test value uses limbs [1, 2, 3, 4], which should be:
```
// Human readable: 25108406941546723056364004793593481054836439088298861789185
1 // 0..8
+ (2 << 64) // 8..16
+ (3 << 128) // 16..24
+ (4 << 192) // 24..32
```

I also just added lines in the test that call `to_le_bytes()` to sanity check.

Let me know if I misunderstood how the limbs are put together and I can fix it. Thank you.

For reference: https://borsh.io/ (see Integers section of Specialization).

rtrombone:
> thank you. does borsh serialize fixed-sized integers as fixed-length bitstrings? i.e. is a u64 always exactly 8 bytes?
> 
> My concern with the current implementation is that non-standard sized `Uint` would currently be serialized at sizes other than their bit size rounded to the nearest byte. E.g. a 7 byte integer like `Uint<56, 1>` would be serialized as 8 bytes. Is this reasonable for Borsh, or should we use try to encode the minimal number of bytes?

You're right. This was an oversight on my part and tried to be clever with iterating over the limbs.

Maybe I should just do to_le_bytes and serialize that. I'll need to look into how to_le_bytes works under the hood for `Uint` to make sure this is kosher.

rtrombone:
I couldn't find a better way than to use a macro to implement `BITS` to `BYTES` relationship. Maybe at some point in the future when we can use `Self::BYTES` this can be generalized.

prestwich:
thanks, let me spend some time on this early next week and i'll see if i can figure out an elegant solution :)

prestwich:
Would you mind checking out my branch [here](https://github.com/recmo/uint/pull/417)?  If you think the implementation is correct, I can push it onto this branch and get this merged soon :)

PR #421, opened 2024-12-24
fix: don't trim when serializing borsh

### LINKED ISSUE #420: Borsh implementation incorrectly serializes uints with trailing zeroe bytes
Introduced by me in #416

If the LE byte repr is `[1, 0]` or any other byte repr with trailing 0s, those zeroes will not be serialized. The deserializers relies on the uint being fixed size when deserializing

prestwich:
closes #420

## Motivation

bug in my impl of #416. 

cc @rtrombone this will cause structs containing uints with trailing 0 bytess that are still within the fixed byte size of the integer to serialize incorrectly

## Solution

don't trim, always write the exact byte size

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


Evalir (approved):

prestwich:
closing in favor of #422

PR #426, opened 2025-01-16
chore(deps): unpin proptest

DaniPopes:
I don't see why it was pinned with `=`. This blocks automatic updates if you already have proptest 1.6 (very likely), as cargo would rather not update ruint than update it and have to downgrade proptest.


prestwich (approved):

prestwich:
just needs the cargo deny addresed

DaniPopes:
Done

PR #431, opened 2025-02-18
fix: swap bytes for `as_le_bytes` in big endian world

pcw109550:
## Motivation

Was tweaking https://github.com/op-rs/kona which has this repo as a dependency. Compiled kona down to mips64 which is big endian, and spotted this bug while comparing the results with little endian arch.

Minimal reproduction:
```rust
use alloy_primitives::U256;
use alloc::vec::Vec;
use alloy_rlp::Encodable;
...
    let mut out = Vec::<u8>::new();
    U256::from(0x12345678).encode(&mut out); // rlp encodes num
```
When we print `out` variable...
- in little endian, `[132, 18, 52, 86, 120]`, which is correct
- in big endian, `[132, 72, 44, 106, 30]`, which is wrong.
both results must be identical. 

## Solution

Must swap bytes for big endian architectures, not reversing bits. 

## PR Checklist

- [ ] Added Tests
    - Unfortunately, no. We may add test suite or ci pipeline for big endian targets
- [ ] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unpin proptest ([#426])
 - Update documentation related to `Uint::byte` and knuth divison ([#429])
 - add `Uint::checked_byte(idx: usize) -> Option<u8>` ([#429])
+- fix: swap bytes for `as_le_bytes` in big endian world ([#431])
 
 [#416]: https://github.com/recmo/uint/pull/416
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):
Right, sorry about that, we don't test the library with big endian so there are likely more of these bugs

prestwich (approved):

prestwich:
i can cut a release later today

prestwich:
this was included in v1.13.0

PR #437, opened 2025-02-26
chore: clippy

DaniPopes:


prestwich (approved):
how should we set a reminder to unpin the clippy version?

PR #438, opened 2025-02-26
fix: `checked_byte` bounds check and make it const

DaniPopes:
`byte_len` returns the dynamic byte length. Use `Self::BYTES` instead.


prestwich:
waiting on #437

PR #441, opened 2025-02-27
feat: constify bitwise and shift functions

DaniPopes:
- Adds `const fn <op>` for all bitwise not, and, or, xor
- Makes `is_zero`, `*shl`, `*shr`, `rotate*`, `count*`, const

WIP: `is_zero` actually performs worse because we have to manually
implement equality, and I cannot get the compiler to generate optimal
code.


prestwich (approved):

prestwich:
just needs changelog

DaniPopes:
@prestwich ready

shekhirin (approved):

DaniPopes:
> do we have intuition for performance of const_eq as opposed to eq?

per godbolt in description they should be identical when both are runtime values, but slightly worse if one is compile-time known like in `is_zero`

Issue #440, opened 2025-02-27
Add `random*` methods on `Uint`


DaniPopes:
In `alloy_primitives::FixedBytes` we have `::random()`, `::randomize()` etc https://github.com/alloy-rs/core/blob/047af040dc4a4c76c3ba0fa1b6e60ba295762937/crates/primitives/src/bits/fixed.rs#L367-L444

Implement these in `rand` support in the same style, treating the `limbs` array as a byte array instead (`[u8; _]`).

The `try_*` methods aren't necessary.

PR #444, opened 2025-02-28
feat: add `random`, `random_with`, `randomize`, `randomize_with` methods

DaniPopes:
Closes https://github.com/recmo/uint/issues/440.


src/support/rand.rs
```diff
@@ -12,15 +12,62 @@ use rand::{
...
+    #[cfg(feature = "std")]
+    pub fn randomize(&mut self) {
+        self.randomize_with(&mut rand::thread_rng());
+    }
+
+    /// Fills this [`Uint`] with the given random number generator.
+    #[inline]
+    #[doc(alias = "randomize_using")]
+    pub fn randomize_with<R: rand::RngCore + ?Sized>(&mut self, rng: &mut R) {
+        if Self::SHOULD_MASK {
```
prestwich, line 63: why is this check necessary? if the u64 randomization algo is uniform, then every bit has equal probability of being set and unset, so masking . so can't you simply fill the full width and always mask?

DaniPopes: yeah i guess that works too

src/support/rand.rs
```diff
@@ -7,20 +7,61 @@
...
+    /// generator.
+    ///
+    /// This is currently [`rand::thread_rng`].
+    #[inline]
+    #[must_use]
+    #[cfg(feature = "std")]
+    pub fn random() -> Self {
+        // SAFETY: `uint` is only accessible after random initialization.
+        #[allow(clippy::uninit_assumed_init)]
+        let mut uint = unsafe { core::mem::MaybeUninit::<Self>::uninit().assume_init() };
```
prestwich, line 32: why unsafety here instead of `Uint::ZERO`?

DaniPopes: uses undefined memory instead of zeroing it

prestwich (approved):

PR #448, opened 2025-03-05
feat: add `const ONE`

DaniPopes:


src/from.rs
```diff
@@ -123,6 +123,15 @@ impl fmt::Display for ToFieldError {
 }
 
 impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
+    pub(crate) const fn const_from_u64(x: u64) -> Self {
```
prestwich, line 126: i know this is not for public consumption, but can you document the behavior when the u64 is out of range?

src/modular.rs
```diff
@@ -84,13 +84,13 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
     #[inline]
     #[must_use]
     pub fn pow_mod(mut self, mut exp: Self, modulus: Self) -> Self {
-        if modulus.is_zero() || modulus <= Self::from(1) {
+        if modulus.is_zero() || modulus <= Self::ONE {
```
prestwich, line 87: looking at this check, isn't it redundant?

if `x == 0 || x <= 1`

prestwich (changes requested):

prestwich (approved):

PR #449, opened 2025-03-05
feat: constify mul*

DaniPopes:
Based on #441


prestwich:
closing as stale. can reopen if we want to

PR #457, opened 2025-05-20
chore: add .DS_Store to gitignore

Rjected:
## Motivation

.DS_Store should be in the gitignore

## Solution

put .DS_Store in the gitignore

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


DaniPopes:
You should just add it to a global gitignore file

prestwich:
@DaniPopes is right that it's better hygiene to enforce locally, but we already have (e.g.) .vscode and other dirs in, so i don't see a problem here

prestwich (approved):

PR #462, opened 2025-05-21
bench: add cmp.rs benchmarks

DaniPopes:
Clean up common code in benchmarks and add cmp.rs

### DIFF
--- a/src/const_for.rs
+++ b/src/const_for.rs
@@ -56,7 +56,7 @@ macro_rules! const_for {
         $crate::const_for!($C in [1, 2, 63, 64, 65, 127, 128, 129, 256, 384, 512, 4096] $x);
     };
     ($C:ident in BENCH $x:block) => {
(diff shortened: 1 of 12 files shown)


src/const_for.rs
```diff
@@ -56,7 +56,7 @@ macro_rules! const_for {
         $crate::const_for!($C in [1, 2, 63, 64, 65, 127, 128, 129, 256, 384, 512, 4096] $x);
     };
     ($C:ident in BENCH $x:block) => {
-        $crate::const_for!($C in [0, 64, 128, 192, 256, 384, 512, 4096] $x);
+        $crate::const_for!($C in [64, 128, 192, 256, 384, 512, 4096] $x);
```
prestwich, line 59: reason for deletion?

DaniPopes: no point in benchmarking a zst, all U0 benchmarks are noops

prestwich (approved):

PR #464, opened 2025-05-21
perf: clean up div, cmp

DaniPopes:
- Remove impossible panics and reduce size of possible panics (div by 0) from div dispatching code to make it more amenable for inlining
- Fix cmp for slices of different lengths; improve comments


prestwich:
blocked by #461

prestwich (approved):

PR #468, opened 2025-05-24
Add Codspeed workflow

recmo:


prestwich (approved):

prestwich:
we can merge here and then merge the improvements from #461

PR #472, opened 2025-05-27
perf: add special case for U256 and limb-specific shift left

Rjected:
## Motivation

Seems like other clients optimize shifts for U256, when the rhs would cause us to only shift limbs. Trying this out here in case it's faster

## Solution

This adds special cases for U256, as well as a limb-by-limb implementation.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


src/bits.rs
```diff
@@ -305,12 +305,51 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
...
+            let mut i = 0;
+            while i < LIMBS - limbs {
+                self.limbs[i] = self.limbs[i + limbs];
+                i += 1;
+            }
+            while i < LIMBS {
+                self.limbs[i] = 0;
+                i += 1;
+            }
+            return (self.masked(), !self.const_is_zero());
```
DaniPopes, line 350: overflow here is always false

src/bits.rs
```diff
@@ -305,12 +305,65 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
...
+            return (self.masked(), false);
+        }
+
         let (limbs, bits) = (rhs / 64, rhs % 64);
         if limbs >= LIMBS {
             return (Self::ZERO, !self.const_is_zero());
         }
 
+        // if there are no leftover bits, shift over just the limbs
+        if bits == 0 && LIMBS == 4 {
```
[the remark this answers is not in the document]
Rjected: interesting, removed

src/bits.rs
```diff
@@ -305,12 +305,39 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
     /// the shift is larger than `BITS` (which is IMHO not very useful).
     #[inline]
     #[must_use]
-    pub const fn overflowing_shl(self, rhs: usize) -> (Self, bool) {
+    pub const fn overflowing_shl(mut self, rhs: usize) -> (Self, bool) {
+        if rhs == 0 {
+            // if there are no limbs and no bits, return self
+            return (self.masked(), false);
```
DaniPopes, line 311: no need to mask an already valid uint; only when you modify the raw limbs is this needed

PR #477, opened 2025-05-29
perf: improve conversion to primitive integers

DaniPopes:
Make the overflow check faster, and branchless for smaller BITS.

These two functions have now identical codegen:
```rust
type T = <integer>;
pub fn u256_to(x: U256) -> Option<T> {
    T::try_from(x).ok()
}

pub fn u256_to_manual(x: U256) -> Option<T> {
    let x = x.as_limbs();
    if (x[1] | x[2] | x[3]) == 0 {
        x[0].try_into().ok()
    } else {
        None
    }
}
```

Whereas before that was only the case for `u64`/`usize` (64 bit unsigned).


src/from.rs
```diff
@@ -425,15 +437,15 @@ impl<const BITS: usize, const LIMBS: usize> TryFrom<u128> for Uint<BITS, LIMBS>
...
             return Self::try_from(value as u64)
                 .and_then(|n| Err(ToUintError::ValueTooLarge(BITS, n)));
         }
         let mut limbs = [0; LIMBS];
         limbs[0] = value as u64;
         limbs[1] = (value >> 64) as u64;
-        if Self::LIMBS == 2 && limbs[1] > Self::MASK {
-            limbs[1] %= Self::MASK;
+        if LIMBS == 2 && limbs[1] > Self::MASK {
+            limbs[1] &= Self::MASK;
```
DaniPopes, line 448: minor drive-by fix: `&` instead of `%`

PR #478, opened 2025-05-29
Avoid allocations in fmt machinery

shamatar:
But still keep it under (different) feature flag

## Motivation

Use of formatting in no_std environments (or environments where `Global` allocator is not desired)

## Solution

Use stack-allocated buffer instead, with clear upper bound on number of possible limbs used for formatting

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Updated the changelog


shamatar:
Sure, I removed the feature gate

PR #483, opened 2025-06-09
feat: add rkyv support

lightsing:
## Motivation

This change adds support for the [rkyv](https://github.com/rkyv/rkyv) serialization framework for the Uint<BITS, LIMBS> type. rkyv enables zero-copy deserialization, making it ideal for performance-critical applications such as memory-mapped files, shared memory, or inter-process communication (IPC).

The project already supports other serialization formats like serde and rlp, so supporting rkyv extends compatibility with a high-performance, archive-based serialization format.

## Solution

This PR introduces a new module, support::rkyv, which implements the necessary traits for rkyv support:
- Archive
- Serialize
- Deserialize
- CheckBytes
- Portable

It also updates Cargo.toml to include rkyv as an optional dependency and exposes the corresponding feature flag. The implementation includes roundtrip property-based tests using proptest to validate the correctness of (de)serialization and archived access.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support [rkyv](https://github.com/rkyv/rkyv) ([#483])
 - Added `bigdecimal` support ([#486])
+- `PartialEq` and `PartialOrd` implementations for primitive integers; minor breaking change for type inference ([#491])
 
 ### Fixed
 
@@ -21,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#476]: https://github.com/recmo/uint/pull/476
 [#483]: https://github.com/recmo/uint/pull/483
 [#486]: https://github.com/recmo/uint/pull/486
+[#491]: https://github.com/recmo/uint/pull/491
 
 ## [1.15.0] - 2025-05-22
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,6 @@ license.workspace = true
 homepage.workspace = true
 repository.workspace = true
 
-[lib]
-bench = false
-
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = [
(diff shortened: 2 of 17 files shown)


prestwich:
I'm not opposed to supporting this, however, i do not know that I can commit to maintaining this across rkyv versions. Given that rkyv is pre-1.0, we would need to differentiate support at each rkyv minor version. How stable is the rkyv interface? Is rkyv@1.0 planned? would this be better homed in rkyv until rkyv is 1.0?

lightsing:
> i do not know that I can commit to maintaining this across rkyv versions.

The rkyv `Archive`, `Deserialize`, `Serialize` style did not change a lot, at least did not change since 0.4 (when they added `Deserialize` and `Serialize`).
A better way is to derive those traits but as I saw other support mod in this crate did not use this approach, so I manually implement it.

> we would need to differentiate support at each rkyv minor version.

or we can only support the latest stable version.

> Is rkyv@1.0 planned? would this be better homed in rkyv until rkyv is 1.0?

As far as I know, they don't have plan to make 1.0 ver.

thx.

lightsing:
It would be easier to maintain like this, but I don't know whether you will accept this?

```
#[cfg_attr(feature = "rkyv", derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize))]
#[repr(transparent)]
pub struct Uint<const BITS: usize, const LIMBS: usize> {
    limbs: [u64; LIMBS],
}
```

see #484

prestwich:
> > we would need to differentiate support at each rkyv minor version.
> 
> or we can only support the latest stable version.

this project's policy is that breaking changes in supported deps qualify as semver major for ruint releases (except in special circumstances like security issues). I.e. _dropping_ support for rkyv@0.8 in favor of rkyv@0.9 would be semver major, and require a ruint@2 release, as crates using ruint's rkyv feature would now fail to compile

prestwich:
> > > > we would need to differentiate support at each rkyv minor version.
> > > 
> > > 
> > > or we can only support the latest stable version.
> > 
> > 
> > this project's policy is that breaking changes in supported deps qualify as semver major for ruint releases (except in special circumstances like security issues). I.e. _dropping_ support for rkyv@0.8 in favor of rkyv@0.9 would be semver major, and require a ruint@2 release, as crates using ruint's rkyv feature would now fail to compile
> 
> Would it be acceptable using feature gate like `rand_08`/`rand_09`?

Yes, that'd be fine

prestwich:
I apparently cannot update the branch for this PR. Please run `cargo +nightly fmt` then merge main and update the PR branch with the 2 new commits

prestwich:
closing the loop, this has been released in v1.16.0

PR #485, opened 2025-06-12
perf: improve, specialize leading_zeros

DaniPopes:
specialize for <= u256

I couldn't figure out a way to have a branchless clz directly, so I opted to specialize for small bit widths instead by using u64/u128 primitives


src/bits.rs
```diff
@@ -155,13 +155,24 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
     #[inline]
     #[must_use]
     pub const fn leading_zeros(&self) -> usize {
+        let fixed = Self::MASK.leading_zeros() as usize;
```
prestwich, line 158: this could be a computed const?

prestwich: it doesn't matter for perf in this context, it does matter for potential reuse by other fns or downstream users. the question is whether it's more intuitive and legible to access this as `Self::MASK.leading_zeros()` or as `Self::MASK_LEADING_ZEROS`. I'm not convinced it's better either way (especailly as `MASK` is not exactly a clear constant name on its own)

PR #487, opened 2025-07-05
chore: add fmt benches

DaniPopes:


DaniPopes:
They run instantly in codspeed but maybe too many? wdyt @Rjected @prestwich

PR #489, opened 2025-07-06
perf: specialize and use subtraction to implement PartialOrd

DaniPopes:
Makes <, <=, >, >= operators branchless by delegating to `lt`, implemented with borrowing_sub chain (a - b produces the overflow boolean for free in the CPU flags)

Currently codegen for the operators is not optimal due to https://github.com/rust-lang/rust/issues/143517


prestwich:
closing as stale

PR #491, opened 2025-07-06
feat: implement PartialEq, PartialOrd for primitive integers

DaniPopes:
## Motivation

This is a minor breaking change to inference as you can see from the changed tests. This is generally acceptable if it doesn't have that big of an impact in practice.

## Solution

Implement `Partial{Eq,Ord}<$int> for Uint`.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/src/cmp.rs
+++ b/src/cmp.rs
@@ -15,6 +15,58 @@ impl<const BITS: usize, const LIMBS: usize> Ord for Uint<BITS, LIMBS> {
     }
 }
 
+/// Implements `PartialEq` and `PartialOrd` for `Uint` and primitive integers.
+///
+/// This intentionally does not use `<$t>::try_from` to avoid unnecessary
+/// checks for non-limb-sized primitive integers.
+macro_rules! impl_for_primitives {
+    ($($t:ty),* $(,)?) => {
+        $(
+            impl<const BITS: usize, const LIMBS: usize> PartialEq<$t> for Uint<BITS, LIMBS> {
+                #[inline]
+                #[allow(unused_comparisons)] // Both signed and unsigned integers use this.
+                #[allow(clippy::cast_possible_truncation)] // Unreachable.
+                fn eq(&self, &other: &$t) -> bool {
+                    (other >= 0) & (if <$t>::BITS <= u64::BITS {
+                        u64::try_from(self).ok() == Some(other as u64)
+                    } else {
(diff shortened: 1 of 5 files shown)


src/support/serde.rs
```diff
@@ -171,12 +171,12 @@ mod tests {
             const LIMBS: usize = nlimbs(BITS);
             proptest!(|(value: Uint<BITS, LIMBS>)| {
                 let serialized = serde_json::to_string(&value).unwrap();
-                let deserialized = serde_json::from_str(&serialized).unwrap();
+                let deserialized = serde_json::from_str::<Uint<BITS, LIMBS>>(&serialized).unwrap();
```
prestwich, line 174: so to check my understanding, inference changed because the `serde_json::from_str` (and the other deser methods) must produce some type `T` that the `assert_eq!` then accepts on the next line. This `T` now has multiple candidates, whereas previously it had only 1 candidate (`Uint<Bits, Limbs>`), right?

I think this is acceptable breakage for our semver policy, as it is rare and can be trivially fixed by adding a type annotation

prestwich (approved):

DaniPopes:
I patched foundry and reth locally to use this, no errors encountered

src/support/serde.rs
DaniPopes: that's right, it is mentioned here as well https://std-dev-guide.rust-lang.org/breaking-changes/new-trait-impls.html#inference-breaks-when-a-second-generic-impl-is-introduced

src/cmp.rs
```diff
@@ -15,6 +15,58 @@ impl<const BITS: usize, const LIMBS: usize> Ord for Uint<BITS, LIMBS> {
...
+                #[inline]
+                #[allow(unused_comparisons)] // Both signed and unsigned integers use this.
+                #[allow(clippy::cast_possible_truncation)] // Unreachable.
+                fn partial_cmp(&self, other: &$t) -> Option<Ordering> {
+                    if *other < 0 {
+                        return Some(Ordering::Greater);
+                    }
+
+                    if <$t>::BITS <= u64::BITS {
+                        let Some(self_t) = u64::try_from(self).ok() else {
```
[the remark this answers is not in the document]
DaniPopes: copied from above

PR #509, opened 2025-07-26
chore: clean up algorithms::ops

DaniPopes:
- make names consistent
- use existing carrying_add and borrowing_sub.


DaniPopes:
Functions in `mod algorithms` are not algorithms and dont respect semver

prestwich:
I see that in the mod docs, but it's not gated by the `unstable` feature described in the README (which does not seem to actually exist). Let's clean up while we're here?

Let's either a) add an `unstable` feature and use it to gate algorithms, or b) add a note to the readme that algorithms is unstable and remove the erroneous reference to the `unstable` feature

DaniPopes:
I'll open another PR for that

DaniPopes:
https://github.com/recmo/uint/pull/511

prestwich (approved):

PR #512, opened 2025-07-27
perf: elide panics in algorithms

DaniPopes:
See individual commits.


src/algorithms/div/mod.rs
```diff
@@ -65,6 +65,7 @@ pub fn div(numerator: &mut [u64], divisor: &mut [u64]) {
 
     if super::cmp(numerator, divisor).is_lt() {
         // Numerator is smaller than the divisor: (q, r) = (0, numerator)
+        assume!(numerator.len() <= divisor.len()); // `a < b` implies `a.len() <= b.len()`.
```
prestwich, line 68: true only after trimming most-significant 0s

prestwich (approved):

Issue #502, opened 2025-07-21
tracking: 2.0


DaniPopes:
I think 2.0 is pretty far away since we're blocked on generic const expressions (GCE), but I think it's worth gathering and discussing all the ideas in one place.

- [ ] Remove `Bits` type
  - This is a thin wrapper around Uint which does not have any extra functionality
- [ ] Remove second `Uint` generic (`LIMBS`), blocked by GCE
  - It could also be a non-breaking change by using a default value
- [ ] Remove support for `BITS == 0` and possibly `BITS == 1`
  - I don't see the point of U0, maybe U1 too. It only causes unnecessary cruft throughout the codebase, where we check for `BITS == 0` and have a special case path because we have 0 limbs
- [ ] Represent Uint internally using machine word-sized Word/DoubleWord type aliases instead of hardcoded u64
  - https://github.com/recmo/uint/issues/371
  - possibly also configurable with compile time features, not just target word size
- [ ] Align parsing behavior with standard library
  - https://github.com/recmo/uint/issues/348
- [ ] Make `as_le_slice` and similar methods return an array of `[u8; Self::BYTES]`
- [ ] Remove support for old crates support features
  - ssz, ark-ff-0.3 ...
- [ ] Resolve the overflow vs panicking arithmetic question: #408, #522

cc @mattsse @gakonst @recmo @prestwich

prestwich:
I'd like to consider 
- [ ] an interface review. do our function names and behaviors conform to those of the std library and other integer libraries
- [ ] variable length uints

recmo:
> variable length uints

After messing with this I've found that something like

```rust
Uint<Limbs: DerefMut<Target = [u64]>> {
   limbs: Limbs
}
```

strikes a nice balance between simplicity and generality.

prestwich:
Sketched out a bit. Is this what you were thinking?

```
struct Uint<Limbs: DerefMut<Target = [Word], const BITS: usize>
```

and then variable-lenght are

```
trait VarLengthWordArray { ... }
impl VarLengthWordArray for Vec<Word> { ... }

trait VarLengthUint { ... }
impl<T, const BITS: usize> VarLengthUint for Uint<T, BITS> where T: VarLengthWordArray { ... }
```

prestwich:
having BITS doesn't really make sense until const generic math support is significantly better. this has got to be a parallel type with an unfixed bit size

PR #524, opened 2025-09-07
Enhanced f64 -> Uint conversion

Pzixel:
## Motivation

I'm using f64 conversions a lot, and when I've checked the crate impl I've seen a lot of expensive computations. I've also seen the TODO that the impl isn't exactly correct so I decided to take a look if I can make it better. 

## Solution

I just took the Rust's STD int<->float conversion code [from here](https://github.com/rust-lang/compiler-builtins/blob/f4c7940d3b13ec879c9fdc218812f71a65149123/src/float/conv.rs#L163) and adapted it pretty much directly.

Note that this is a very initial impl, if this is something you want to have I can add tests, docs and anything else that is needed

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support for [bincode 2](https://github.com/bincode-org/bincode) ([#516])
 - Introduce `ark-ff-05` feature flag for conversion to `ark-ff@0.5` types ([#526])
 
 ### Changed
@@ -20,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reimplemented `TryFrom<f64>` for `Uint` to speed it up, fixing edge cases and removing `std` requirements ([#524])
 
 [#503]: https://github.com/recmo/uint/pull/503
+[#516]: https://github.com/recmo/uint/pull/516
 [#526]: https://github.com/recmo/uint/pull/526
 [#533]: https://github.com/recmo/uint/pull/533
 [#524]: https://github.com/recmo/uint/pull/524
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ ark-ff-03 = { version = "0.3.0", package = "ark-ff", optional = true, default-fe
 ark-ff-04 = { version = "0.4.0", package = "ark-ff", optional = true, default-features = false }
 ark-ff-05 = { version = "0.5.0", package = "ark-ff", optional = true, default-features = false }
 bigdecimal = { version = "0.4", optional = true, default-features = false }
+bincode-2 = { version = "2", package = "bincode", optional = true, default-features = false }
 bn-rs = { version = "0.2", optional = true, default-features = true }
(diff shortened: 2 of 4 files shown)


DaniPopes (approved):

Pzixel:
I have an immediate question: for example consider `Uint::<1,1>::try_from(2.0)`. The new impl returns `ToUintError::ValueNegative(1,1)`. The old impl however returns `ToUintError::ValueNegative(1,0)`. It feels to me like the `wrapped` value should be 1 in this case. Since the comment on the function was `// TODO: Correctly implement wrapping.` this change is probably allowed?

Pzixel:
I have one question though: the current implementation saturates out of range values, whether the old impl does wrapping. Do we want to keep this behaviour? I know it's a breaking change, but'm not sure if anyone actually used this and if it's better to bring the impl closer to an STD one or keep the existing behaviour

prestwich:
doing a survey of downstream packages seems reasonable but very time-consuming. or we could bump this to be part of #502

Pzixel:
Oh, that actually seems very reasonable.

That issue mentions that 2.0 is blocked by the nightly feature, that won't probably be merged in the next couple of years. I will reimplement the old behaviour for the errors. So we will keep the benefits for the happy path, and the error path will be the same as it is now. 

The only problem is that I want to have a no_std version, so I need to reimplement this without using std features that are currently in use. I'l figure it out soon.

prestwich:
>  I will reimplement the old behaviour for the errors. So we will keep the benefits for the happy path, and the error path will be the same as it is now.

okay I appreciate that. it's much easier to justify a semver minor or patch release. please let me know if you end up blocked in any way

prestwich:
@DaniPopes the old one also relies on `std` so adaptation is needed regardless

prestwich:
> This value is beyond the f64 precision, so we kinda are allowed to have this change, but I wanted to confirm you're agreeing with that.

this seems acceptable to me :)

> I also think we need to separate perf report into Happy path/ Erroneous path, so we can see more clearly how PR affects the codebase. Also the 4096 bench seems off - does it really take just 50 us to compute?

it's possible we haven't sufficiently black boxed the bench functions. feel free to include some bench modifications / new benches in this PR. if you don't want to write them, i can do it within the next few days

Pzixel:
There probably remains a single question: the new impl uses `round-to-nearest, ties-to-even` logic. So the  Uint::<1, 1>(1625960292148408.5) value rounds to 1625960292148408 => ValueTooLarge(0), whilte the older was 1625960292148408.5 => 1625960292148409 => ValueTooLarge(1). If this is an okay change then this is the final commit.

prestwich:
> There probably remains a single question: the new impl uses `round-to-nearest, ties-to-even` logic (because the Rust impl uses it). So the Uint::<1, 1>(1625960292148408.5) value rounds to 1625960292148408 => ValueTooLarge(0), whilte the older was 1625960292148408.5 => 1625960292148409 => ValueTooLarge(1). If this is an okay change then this is the final commit.

switching to baker's rounding seems acceptable to me. i will merge shortly, and also tweet about this. If nobody gets annoyed within a few days, we'll publish this in the next patch version

prestwich (approved):

PR #525, opened 2025-09-08
bench: bench from f64 and f32

prestwich:
related to #524 

## Motivation

Add a bench to the try_from impls for f32 and f64 so that we can roughly understand the impact of #524


prestwich:
clsoing in favor of #531

PR #534, opened 2025-09-17
Add scale-info support

pgherveou:
## Motivation

Add support for scale-info https://docs.rs/crate/scale-info

## Solution

Added support in support/scale_info via the scale-info feature flag.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/src/support/scale_info.rs
+++ b/src/support/scale_info.rs
@@ -1,0 +1,31 @@
+//! Support for the [`sale-info`](https://crates.io/crates/scale-info) crate.
+
+#![cfg(feature = "scale-info")]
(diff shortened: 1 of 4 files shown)


src/support/scale_info.rs
```diff
@@ -0,0 +1,29 @@
...
+                ::core::module_path!(),
+                &[],
+            ))
+            .composite(
+                scale_info::build::Fields::unnamed()
+                    .field(|f| f.ty::<[u64; LIMBS]>().type_name("[u64; LIMBS]")),
+            )
+    }
+}
+
```
prestwich, line 24: should be in a `#[cfg(test)] mod test;`

PR #535, opened 2025-09-19
Add Uint -> f64 conversion

Pzixel:
## Motivation

Adds more conversions, completing #524 , allowing f64 <-> Uint conversions without the STD feature, hopefully speeding them up as well

## Solution

Again I used rust built-in code as an example as well as an article https://blog.m-ou.se/floats/

There are no benchmarks yet, I understand they need to get into the `main` branch first

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Updated the changelog

### DIFF
--- a/src/from.rs
+++ b/src/from.rs
@@ -768,7 +768,6 @@ impl<const BITS: usize, const LIMBS: usize> TryFrom<&Uint<BITS, LIMBS>> for u128
 
 // Convert Uint to floating point
 
-#[cfg(feature = "std")]
 impl<const BITS: usize, const LIMBS: usize> From<Uint<BITS, LIMBS>> for f32 {
     #[inline]
     fn from(value: Uint<BITS, LIMBS>) -> Self {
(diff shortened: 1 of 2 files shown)


src/from.rs
```diff
@@ -940,6 +1012,88 @@ mod test {
         });
     }
 
+    #[test]
+    fn all_floats_are_representable() {
+        const MAX_SAFE_INTEGER: u64 = (1 << f64::MANTISSA_DIGITS) - 1;
+        proptest!(|(value in 0..MAX_SAFE_INTEGER)| {
```
DaniPopes, line 1018:
```suggestion
        proptest!(|(value in 0..=MAX_SAFE_INTEGER)| {
```

Pzixel:
Commit `d30adf0f5c4511731cffd3f8bd0b3855baccf34f` can be cherry-picked to `main` to measure the difference

PR #540, opened 2025-09-26
Update primitive-types to v0.14

Teebor-Choka:
This pull request updates the version of the `primitive-types` dependency in `Cargo.toml` to improve compatibility and access to newer features.

Dependency updates:

## Motivation
* Upgraded `primitive-types` from version `0.12` to `0.14` in `Cargo.toml` to ensure the project uses the latest features and bug fixes.

The used version is obsoleted and causes the dependents to have multiple parallel 

## Solution
Bump the `primitive-types` dependency.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,7 +93,7 @@ parity-scale-codec = { version = "3", optional = true, default-features = false,
     "derive",
     "max-encoded-len",
 ] }
-primitive-types = { version = "0.12", optional = true, default-features = false }
+primitive-types = { version = "0.14", optional = true, default-features = false }
 proptest = { version = "1", optional = true, default-features = false, features = [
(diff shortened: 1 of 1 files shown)


prestwich:
our semver policy considers breaking updates of deps from which we impl traits to be breaking upgrades. I.e. this would be considered semver major under our policy, because it would remove support for 0.12 thereby breaking downstream crates that rely on 0.12 support. As such, this would be delayed for #502 

To **add** primitive-types support to 0.14, you can follow the pattern we've established for ark-ff. This pattern is strictly additive, and would be in the next semver minor

https://github.com/recmo/uint/pull/526

PR #542, opened 2025-10-06
Refactor: refactor most_significant_bits to use branchless shift operations

malik672:
This refactor achieves the same safety using a double-shift pattern and based on godbolt we can save 655 cycles


prestwich:
regardless of whether the bench exists, i want ci to pass before merge 😅

prestwich:
also a clippy to resolve :)

prestwich:
#543 will fix the benches and add a bench for this method

prestwich:
regressions are unrelated, probably related to a new rust version?

PR #547, opened 2025-10-18
add shift benchmark

malik672:
benchmark for shift operation to unblock #546
@prestwich


prestwich:
closing as stale, can revisit if this is still valuable

PR #570, opened 2026-04-21
chore: bump MSRV to 1.90

DaniPopes:
Replace the manual const reverse loop in `to_be_bytes` with `bytes.reverse()`, which is const-stable as of Rust 1.90.


prestwich (approved):

prestwich:
should we prep a minor bump?

PR #574, opened 2026-04-23
perf: optimize shift implementations

DaniPopes:
Optimize shift left/right for better codegen:

- Replace the two-step carry trick `(x >> (64 - bits - 1)) >> 1` with `x.unbounded_shr(64 - bits)` in the generic loop, exposing the funnel shift pattern to LLVM.
- Add specialized fast paths for u64, u128, and u256 in `wrapping_shl`/`wrapping_shr` via the `as_primitives!` macro.
- The u256 path decomposes into two u128 halves with branchless mask blending, producing `shld`/`shrd` + `cmov` with zero branches in the hot path.

llvm-mca (Zen 4) for `U256::wrapping_shl`:
- 57 instructions, 73 uOps, ~13.1 cycles latency, IPC 4.35
- Compared to native `shl i256`: 36 instructions, 37 uOps, ~15.1 cycles latency, IPC 2.39
- Our version has better latency due to avoiding store-forwarding penalties; native `shl i256` has better throughput due to fewer instructions.


DaniPopes:
only shr benches are impacted by this change, rest is noise: https://github.com/recmo/uint/pull/573#issuecomment-4308363453

prestwich (approved):

PR #575, opened 2026-04-25
feat!: ban BITS == 0

DaniPopes:
This way too many branches for something that is practically useless. This should've never been allowed.

I'm pretty confident nobody is using this. If there is, we can always revert this.


prestwich:
i'm generally in favor, however, this is a 2.0 change. are we planning a 2.0 release and if so, what else do we want in?

prestwich:
related: #502

DaniPopes:
Gonna wait for this

PR #576, opened 2026-04-28
Fix `to_be_bytes` on big-endian platforms

kayabaNerve:
This assumed `u64` would have a little-endian representation and unconditionally call `swap_bytes`.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -157,8 +157,24 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
             let mut limbs_be = [0u64; LIMBS];
             let mut i = 0;
             while i < LIMBS {
-                limbs_be[i] = self.limbs[LIMBS - 1 - i].swap_bytes();
+                #[cfg(target_endian = "big")]
+                {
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):

src/bytes.rs
```diff
@@ -155,10 +155,26 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
...
-                i += 1;
+            #[cfg(target_endian = "big")]
+            {
+                limbs_be = self.limbs;
+            }
+            #[cfg(target_endian = "little")]
+            {
+                let mut i = 0;
+                while i < LIMBS {
+                    limbs_be[i] = self.limbs[LIMBS - 1 - i].swap_bytes();
```
DaniPopes, line 166: i think this needs to be `.to_be()` for the b-e path too. the limbs are in l-e order on both architectures

DaniPopes (changes requested):

src/bytes.rs
kayabaNerve: Ah. Durr. Sorry. Will fix in a moment.

prestwich (approved):

PR #579, opened 2026-04-28
perf: compare using subtraction

DaniPopes:
This resurrects the comparison approach from #489 on top of current main.

It implements limb-slice comparison through a borrow chain over `a - b`, using the accumulated difference to detect equality and the final borrow to detect ordering. `Ord::cmp` and the `PartialOrd` relational operators use the algorithms module directly.

On x86-64, the implementation uses the stable `_subborrow_u64` intrinsic, which lowers to `llvm.x86.subborrow.64` and lets LLVM select the `sub/sbb` chain without the volatile codegen hack. Other targets keep the existing `borrowing_sub` helper and volatile fallback to preserve the borrow-chain shape there.

For U256, the relational operators lower to the same core `cmp/sbb/sbb/sbb/setcc` shape as LLVM `i256 icmp`. LLVM's `llvm.ucmp.i32.i256` / C++ `<=>` lowering for `cmp` uses two borrow chains, while this implementation uses one borrow chain plus an OR-reduction of the subtraction result to detect equality.

The existing primitive specializations for u64 and u128 remain in place. Tests compare `cmp`, `<`, `>`, `<=`, and `>=` against an independent high-to-low limb reference across the existing SIZES set.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2024"
-rust-version = "1.91"
+rust-version = "1.90"
 authors = ["Remco Bloemen <remco@wicked.ventures>"]
 license = "MIT"
 homepage = "https://github.com/recmo/uint"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2024"
-rust-version = "1.90"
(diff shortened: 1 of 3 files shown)


DaniPopes:
Regression is on a single synthetic value, it's pseudo faster because it gets branch predicted on the jump vs constant time comparing all limbs. So it's not actually regressing, confirmed with assembly output.

prestwich (approved):

PR #585, opened 2026-05-08
ci: make Accept gate fail when any required job fails

prestwich:
## Summary

The `Accept` job in `ci.yml` was a single-step `run: "true"` that depended on all the real CI jobs. With default GitHub Actions semantics, when any job in `needs:` fails, dependent jobs are **skipped**, not failed.

GitHub branch protection treats a `skipped` required status check as non-blocking, so the required `Accept` check never reported `failure` even when `test`/`clippy` failed. PR #576 hit exactly this — automerge fired with failing CI because `Accept` was reported as `skipped` rather than `failure`.

This PR runs `Accept` unconditionally (`if: always()`) and explicitly fails when any dep is not `success`, turning it into a real failing required check that branch protection blocks on.

## Test plan

- [ ] Open this PR; confirm `Accept` runs and reports `success` when all jobs pass.
- [ ] Push a commit that breaks a test on a draft PR; confirm `Accept` reports `failure` (not `skipped`) and that branch protection blocks merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,10 @@ jobs:
     if: always()
     needs: [test, feature-checks, clippy, docs, fmt]
     steps:
-      - name: Check all needs succeeded
-        if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') || contains(needs.*.result, 'skipped') }}
-        run: exit 1
+      - name: Decide whether the needed jobs succeeded or failed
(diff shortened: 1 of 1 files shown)


DaniPopes (commented):
can we do this?
https://github.com/alloy-rs/core/blob/2ebfa5b502e535d9864497579ebac7eeae30e27d/.github/workflows/ci.yml#L232-L253

prestwich:
🫡

DaniPopes (approved):

PR #591, opened 2026-05-23
feat: make from_str_radix const

DaniPopes:
Makes `from_str_radix` usable in const contexts. Tweak const_range_for macro to make it have similar codegen to the original range/slice iterators, to preserve performance.


mattsse (approved):
LGTM

PR #594, opened 2026-06-11
chore: group weekly dependabot updates

decofe:
## Summary
- run all Dependabot ecosystems on a weekly schedule
- group dependency updates with a catch-all pattern
- group GitHub Actions/CI updates as `ci-weekly`

## Testing
- validated `.github/dependabot.yml` parses as YAML and every update entry is weekly with a dependency group

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,24 @@
 version: 2
 updates:
-  - package-ecosystem: "cargo"
-    directory: "/"
+  - package-ecosystem: cargo
+    directory: /
     schedule:
-      interval: "weekly"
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

DaniPopes (approved):

prestwich:
while here, is it easy to explicitly exclude optional crates from dependabot resolution? we generally do _not_ want automatic updates of feature crates, as our impls are version-specific and updating a dep version results in semver major bumps in ruint

DaniPopes:
#597, removing cargo updates altogether since almost all dependencies are "feature deps" anyway

PR #595, opened 2026-06-14
feat: add strict_* operations

### LINKED ISSUE #532: Implement `strict_*` operations
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Integer `strict_*` operations are being stabilized: https://doc.rust-lang.org/stable/std/primitive.usize.html#method.strict_add
We should add these variants to mirror the standard library's

**Describe alternatives you've considered**

**Additional context**

0xMars42:
Closes #532

Adds the `strict_*` arithmetic operations to mirror the standard library's (now stabilized) variants. They always panic on overflow, regardless of whether overflow checks are enabled, complementing the existing `checked_*` (returns `Option`), `wrapping_*`, `saturating_*` and `overflowing_*` families.

Added, each next to its `checked_*` counterpart and reusing the existing `overflowing_*` building blocks:

- `strict_add`, `strict_sub`, `strict_neg` (`add.rs`)
- `strict_mul` (`mul.rs`)
- `strict_pow` (`pow.rs`)
- `strict_shl`, `strict_shr` (`bits.rs`)
- `strict_div`, `strict_rem` (`div.rs`)

`const`-ness, `#[inline]`/`#[must_use]` attributes and panic messages follow the standard library and the surrounding code. `strict_shl`/`strict_shr` panic when the shifted-out bits are non-zero, matching ruint's `checked_shl`/`checked_shr` semantics (which intentionally differ from `u64::checked_shl`).

Tests cover the success path and a `#[should_panic]` case per operation, asserting the exact panic message.

### DIFF
--- a/src/add.rs
+++ b/src/add.rs
@@ -31,6 +31,22 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         }
     }
 
+    /// Computes `self + rhs`, panicking if overflow occurred.
+    ///
+    /// # Panics
+    ///
+    /// This function will always panic on overflow, regardless of whether
+    /// overflow checks are enabled.
+    #[inline(always)]
+    #[must_use]
+    #[track_caller]
+    pub const fn strict_add(self, rhs: Self) -> Self {
+        match self.overflowing_add(rhs) {
+            (value, false) => value,
+            _ => panic!("attempt to add with overflow"),
+        }
+    }
+
     /// Computes `-self`, returning [`None`] unless `self == 0`.
     #[inline(always)]
     #[must_use]
@@ -41,6 +57,22 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         }
     }
 
+    /// Computes `-self`, panicking unless `self == 0`.
+    ///
+    /// # Panics
+    ///
+    /// This function will always panic on overflow, regardless of whether
+    /// overflow checks are enabled.
+    #[inline(always)]
+    #[must_use]
+    #[track_caller]
+    pub const fn strict_neg(self) -> Self {
+        match self.overflowing_neg() {
+            (value, false) => value,
+            _ => panic!("attempt to negate with overflow"),
+        }
+    }
+
     /// Computes `self - rhs`, returning [`None`] if overflow occurred.
     #[inline(always)]
     #[must_use]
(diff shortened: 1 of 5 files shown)


prestwich:
Logic looks good. Can you break the tests out into the appropriate modules instead of all in `add.rs`?

0xMars42:
Good call. Moved each strict_* test into the module of its operation: mul in mul.rs, div/rem in div.rs, pow in pow.rs, shl/shr in bits.rs. add/sub/neg stay in add.rs.

prestwich:
Reviewed for conformance to the std `strict_*` methods and pushed two small changes to this branch.

**Conformance:** the 7 arithmetic/div ops (`strict_add`, `strict_sub`, `strict_neg`, `strict_mul`, `strict_pow`, `strict_div`, `strict_rem`) match std exactly — same panic conditions and identical panic messages. `strict_neg` correctly uses unsigned semantics (panics unless `self == 0`), and `strict_div`/`strict_rem` correctly handle only divide-by-zero (no unsigned overflow case). Signature choices (`exp: Self` for pow, `rhs: usize` for shifts, non-`const` div/rem) follow ruint's own conventions rather than std, which is the right call for this crate.

**One behavioral divergence — `strict_shl`/`strict_shr`:** these build on ruint's `overflowing_shl`/`overflowing_shr`, which flag overflow when **non-zero bits are shifted out**, whereas `u64::strict_shl`/`strict_shr` panic when the **shift amount ≥ bit width**. The panic *messages* are identical, so the difference is easy to miss — e.g. `3u64.strict_shr(1)` panics here but returns `1` in std. This is intentional and consistent with the existing `checked_`/`overflowing_` shift family, so no code change — but I added an inline `Note:` cross-referencing `u64::strict_shl`/`strict_shr`, matching the note the neighboring shift methods already carry.

**Changes pushed:**
- `docs:` added the std cross-reference note to `strict_shl`/`strict_shr`.
- Merged latest `main` into the branch.

No correctness issues. LGTM aside from the doc note now added.

prestwich (approved):
Approving — std conformance verified, doc note added, main merged. The shift-semantics divergence is intentional and now documented.

src/add.rs
```diff
@@ -31,6 +31,22 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
...
+    ///
+    /// This function will always panic on overflow, regardless of whether
+    /// overflow checks are enabled.
+    #[inline(always)]
+    #[must_use]
+    #[track_caller]
+    pub const fn strict_add(self, rhs: Self) -> Self {
+        match self.overflowing_add(rhs) {
+            (value, false) => value,
+            _ => panic!("attempt to add with overflow"),
```
DaniPopes, line 46: nit: let's match std by outlining each panic into its own function to reduce callsite size

prestwich:
Heads up: I squashed this branch into a single commit and force-pushed to `feat/strict-ops` to ensure the history is signed (some of the earlier commits were unsigned / signed with a key our verification doesn't recognize).

- All previous commits squashed into one: `feat: add strict_* operations`
- You're preserved as the commit author; it's now signed with a verified key
- Rebased onto the latest `main` (your branch had merged in #600/#601, which are already upstream)
- The file tree is byte-identical to before the squash — no code changes

If you have local work on this branch, please `git fetch` and reset to the new remote tip (`git reset --hard origin/feat/strict-ops`) before continuing to avoid re-introducing the old history.

PR #597, opened 2026-06-21
chore: group weekly GitHub Actions updates

DaniPopes:
Supersedes #594.

This removes the Cargo Dependabot entry and keeps GitHub Actions version updates grouped into a single weekly `ci-weekly` PR, scheduled for Monday at 09:00 UTC.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,48 +1,5 @@
 version: 2
 updates:
-  - package-ecosystem: cargo
-    directory: /
-    schedule:
-      interval: weekly
-      day: monday
-      time: "09:00"
-      timezone: UTC
-    ignore:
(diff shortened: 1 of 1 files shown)


prestwich:
okay. the logic as i understand it is as follows:
- in `ruint` we have only optional dependencies
- optional dependencies are not pinned to specific versions, follow standard Cargo dep resolution
    - this allows dependent crates to freely select semver non-breaking versions
    - **so dependabot is not required for non-breaking dep versions**
- any semver breaking dep update needs to be implemented as a new versioned support file
    - **so dependabot is not allowed for semver breaking dep versions**
- therefore we have no need of dependabot for cargo deps

and ruint-macro has 0 deps

prestwich (approved):

PR #603, opened 2026-07-08
fix: overflowing_shl/overflowing_shr overflow-flag false negatives

prestwich:
## Summary

This PR fixes several correctness bugs in `Uint`'s shift operations and in `byte()`.

### 1. `overflowing_shl`/`overflowing_shr` overflow-flag false negatives

- `overflowing_shl`'s overflow flag missed bits shifted above `BITS` but still inside the top limb (any non-limb-aligned width, e.g. U65, U160), and source limbs wholly discarded when the shift amount is `>= 64`.
- `overflowing_shr`'s overflow flag missed low limbs wholly discarded when the shift amount is `>= 64`.
- Both false negatives propagate to `checked_shl`/`checked_shr`, `strict_shl`/`strict_shr`, `saturating_shl`/`saturating_shr`, and cause `to_base_be` to loop forever on no-alloc builds for non-limb-aligned widths (`checked_shl` incorrectly returns `Some(0)` instead of `None`, so the spigot loop's shrinking condition never terminates).
- The shifted *value* was always correct in both functions — only the overflow flag was wrong.

### 2. Shift-amount truncation in wrapping/overflowing shifts

- The `wrapping_shl`/`wrapping_shr` primitive fast paths (LIMBS in {1, 2, 4}) cast the shift amount `rhs as u32`, reducing it mod 2^32, so shifts `>= 2^32` wrapped instead of producing zero. Fixed with a branchless saturating `usize → u32` cast (`shift_amount`) applied inside each fast-path arm: shifts `> u32::MAX` saturate so `unbounded_sh*` shifts the whole value out (matching the generic path). This lives only in the fast paths — the generic path is `usize` end-to-end and `overflowing_shl` already handles any `rhs` — so no branch is added to the generic path (`wrapping_{shl,shr}/192` stay at their `main` timings).
- `overflowing_shl_big`/`overflowing_shr_big` narrowed the shift amount to `u64` and then cast `rhs as usize`, truncating u64 → u32 on 32-bit targets (e.g. wasm32) and wrapping shift amounts in `[2^32, 2^64)` mod 2^32. Guarded with `rhs >= BITS`, which shifts the whole value out and keeps the surviving `rhs < BITS <= usize::MAX` so the cast can no longer truncate on any target.

### 3. `byte()` panic contract on big-endian targets

- Docs state "Panics if `index >= BYTES`". The little-endian path indexes `as_le_slice()` (length `BYTES`) and panics correctly. The big-endian path indexed `self.limbs[index / 8]`, which only panics for `index >= LIMBS * 8`, so for `BYTES <= index < LIMBS * 8` (sizes with `BITS % 64 != 0`) it silently returned `0` instead of panicking as documented.
- Added an explicit bounds check to the big-endian path so `byte()` upholds its documented contract on all targets.

## Test plan

- [x] `cargo test` (unit + doc tests) passes
- [x] New regression tests `correctness_8_7_2026_overflowing_shl`/`_shr` and `correctness_8_7_2026_overflowing_big` in `src/bits.rs` cover the previously-missed cases (whole-limb shift-out, non-limb-aligned mask boundary, and shift-amount truncation)
- [x] `cargo clippy --tests` clean
- [x] Big-endian path verified to compile via `cargo check --target s390x-unknown-linux-gnu`


src/bits.rs
```diff
@@ -435,6 +446,10 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
     #[inline(always)]
     #[must_use]
     pub const fn wrapping_shl(self, rhs: usize) -> Self {
+        if rhs > Self::BITS {
```
DaniPopes: `unbounded_sh*`:

> If rhs is larger or equal to the number of bits in self, the entire value is shifted out, and 0 is returned.

so i don't think we need this in wrapping_* methods

prestwich: so basically this is a behavior contract issue, suppose `rhs` is `2**32 + 1`. the `as_primitive` match arms will truncate to a `u32` via the `as` cast, which will be `1` and then shift 1 bit. 

this creates a difference in behavior with the generic path, where

`U64(x) >> 2**32 + 1` actually differs from `U257(x) > 2**32 + 1`

we need a check before the `as_primitive` that catches that

prestwich: Or the same check in every as_primitive branch

prestwich: Done in bb4f459 — went with the per-branch approach. The guard only ever existed to stop the `rhs as u32` casts in the `LIMBS ∈ {1,2,4}` fast paths from truncating shifts `>= 2^32`; the generic path runs on `usize` end-to-end and `overflowing_shl` already returns the correct zero for any `rhs`, so it never needed the check.

Replaced the prologue `if rhs > Self::BITS` with a branchless saturating `usize → u32` cast (`shift_amount`, on top of `select_unpredictable_u32`) inside each fast-path arm. For any `rhs > u32::MAX` it saturates to `u32::MAX`, so `unbounded_sh*` shifts the whole value out (matching the generic path) instead of wrapping mod 2^32. No-op the compiler elides on 32-bit targets.

Turns out the prologue branch was hurting more than CodSpeed's ~14% — it was an always-not-taken branch (the bench shifts by `0..=BITS`) that blocked the fall-through into `overflowing_shl` from being optimized. Local A/B/C at size 192 (generic path):

| bench | `main` | this PR before | this PR after |
|---|---|---|---|
| `wrapping_shl/192` | 4.25 ns | 9.13 ns | **4.25 ns** |
| `wrapping_shr/192` | 4.32 ns | 7.63 ns | **4.27 ns** |

Fully back to `main`, and the fast paths are unchanged (saturating cast costs nothing measurable). Note `overflowing_shl/192` stays regressed — that's the discarded-limb carry scan the overflow-flag fix requires, not something this touches.

src/bits.rs
```diff
@@ -1297,4 +1330,90 @@ mod tests {
     fn test_strict_shr_overflow() {
         let _ = crate::aliases::U64::from(1u64).strict_shr(1);
     }
+
+    #[test]
+    fn correctness_8_7_2026_overflowing_shl() {
```
DaniPopes: correctness_8_7_2026 :sob:

src/bits.rs
```diff
@@ -419,9 +441,12 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         let Ok(rhs) = u64::try_from(rhs) else {
             return (Self::ZERO, true);
         };
-        // Rationale: if BITS is larger than 2**64 - 1, it means we're running
-        // on a 128-bit platform with 2.3 exabytes of memory. In this case,
-        // the code produces incorrect output.
+        // Any shift >= BITS shifts the entire value out. Handling it here also
+        // guarantees `rhs < BITS <= usize::MAX`, so the cast below cannot
+        // truncate on 32-bit targets (where `usize` is narrower than `u64`).
+        if rhs >= BITS as u64 {
```
DaniPopes: maybe just do usize::try_from above? and fix the `, true` to `, !self.const_is_zero()`

src/bits.rs
```diff
@@ -552,9 +582,12 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         let Ok(rhs) = u64::try_from(rhs) else {
             return (Self::ZERO, true);
         };
-        // Rationale: if BITS is larger than 2**64 - 1, it means we're running
-        // on a 128-bit platform with 2.3 exabytes of memory. In this case,
-        // the code produces incorrect output.
+        // Any shift >= BITS shifts the entire value out. Handling it here also
```
DaniPopes: same

src/bits.rs
prestwich: don't bother me i'm lazy

src/bits.rs
prestwich: Done in d7cdcba. `usize::try_from` replaces both the u64 narrowing and the `rhs >= BITS` guard — anything that fails the conversion is `> usize::MAX >= BITS`, and anything that survives is handled by `overflowing_shl` itself. Flag on that path is now `!self.const_is_zero()`, with assertions pinning `ZERO << 2^128` → `(ZERO, false)`.

src/bits.rs
prestwich: Same fix applied to `overflowing_shr_big` in d7cdcba.

src/bits.rs
prestwich: Renamed to `regression_*` in d7cdcba.

src/bits.rs
```diff
@@ -533,11 +555,16 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         let bits = bits as u32;
 
         let mut carry = 0;
+
```
DaniPopes:
```suggestion
```
or copy the comment from the other fn

src/bits.rs
```diff
@@ -438,18 +460,18 @@ impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
         as_primitives!(self; {
             u64(x) => {
                 let mut r = Self::ZERO;
-                r.limbs[0] = x.unbounded_shl(rhs as u32);
+                r.limbs[0] = x.unbounded_shl(shift_amount(rhs));
```
DaniPopes, line 468: is this a noop on x64 too?

DaniPopes:
I think codspeed is noise because of new rust release

pw:
Independently ran into the exact bug this PR fixes while differential/boundary-fuzzing an EVM-compatible interpreter that resolves `ruint` 1.19.0 — worth noting the real-world blast radius, since I don't think it's obvious from the PR description alone.

Confirmed root cause matches this PR: `saturating_shl`/`saturating_shr` (and by extension anything built on `overflowing_shl`/`overflowing_shr`) fail to saturate correctly for large shift amounts on non-limb-aligned widths. Boundary-mapped it independently and found the precise trigger condition: wrong whenever `shift mod 2^32 < 256`, for non-negative operands, in the fast-path arms doing the `rhs as u32` truncation. Traced the exact wrong line and confirmed byte-for-byte against predicted output at several boundary points (`2^150`, `2^123`, `2^122`).

Flagging because this isn't just a value-correctness nit for arbitrary-precision math — for anything implementing EVM semantics on top of `ruint` (SAR opcode saturation is spec-defined behavior), this produces a genuine consensus-divergence-shaped bug. It's the kind of thing that's silent until two implementations disagree.

Two asks, since crates.io currently serves 1.19.0 (the vulnerable version) with zero warning:

1. Anything blocking merge beyond the CodSpeed perf comparison, which @DaniPopes already flagged as likely CI noise from a Rust toolchain bump?
2. Once merged/released, would you consider requesting a RUSTSEC advisory (rustsec.org / rustsec/advisory-db)? Right now there's no GHSA or RUSTSEC entry for this, so `cargo audit` gives no signal and a fresh `cargo add ruint` silently resolves the affected version.

Happy to share the independent boundary-test matrix if useful for the regression suite.

DaniPopes:
- comments and CI are not addressed
- in revm this is guarded against https://github.com/bluealloy/revm/blob/46c64ad0e319f3f7da617e8fcfcb1e15c5d17d41/crates/interpreter/src/instructions/bitwise.rs#L137-L141
- in std number shifts anything above bit width is panic/UB, so I don't expect this to have any real impact

src/bits.rs
prestwich: I benched this on my x86 box and checked codegen. for 64/128 bit this is a noop. for 256, there's a small extra asm block to run the `min`. Guessing llvm can't prove its redundant at sizes above 128

```
asm
mov  $0xffffffff,%r10d
cmp  %r10,%r9
cmovb %r9,%r10          ; r10 = min(rhs, u32::MAX), feeds %cl for all shifts
```

on my x86 box, this translates to ~0.3 ns (6%ish) worse perf for this bench

gonna try it with a plain truncate and mask real quick

prestwich: update: i don't think there's another correct configuration that I can find this morning that results in a noop on U256. i'd like to get this in as is

prestwich:
@pw after we release 1.20, I'm fine submitting or having you submit a RUSTSEC advisory

@DaniPopes this is ready for review again

prestwich:
bump @DaniPopes

DaniPopes (approved):

prestwich:
@pw ruint @ 1.20.0 has been released. rustsec advisory has been filed here

https://github.com/rustsec/advisory-db/pull/3087

PR #604, opened 2026-07-08
perf: open-code Uint PartialEq and is_zero on riscv to avoid memcmp libcall

Qumeric:
## Summary

Open-codes `Uint`'s `PartialEq` and `is_zero` on riscv targets only; all other targets keep the derive and the `*self == Self::ZERO` form, unchanged.

On riscv, the derived `[u64; LIMBS]` comparison lowers to a `bcmp`/`memcmp` libcall — the RISC-V backend has no inline-memcmp expansion — and the call overhead (argument setup, length-generic dispatch, byte-at-a-time loop) dwarfs the comparison itself. EVM implementations built on `ruint` hit `==`/`is_zero` in hot interpreter paths millions of times per block on riscv zkVMs.

On riscv this PR replaces them with branchless limb loops:

- `PartialEq::eq`: OR-accumulate the per-limb XORs, then test against zero.
- `is_zero`: OR-accumulate the limbs directly, which also avoids materializing a `ZERO` on the stack just to compare against it.

## Impact

Executing a mainnet Ethereum block in an rv32im zkVM, removing the libcall from these comparisons cut ~3.7% of *all* guest instructions. The cost is spread across many small call sites, so per-site profiles understate it.

## Notes

- Measured on riscv32; riscv64 is included in the gate because the backend lowering is the same (libcall, no inline expansion). Happy to narrow the gate to riscv32 if you'd rather keep it strictly to measured targets.
- On riscv, `Uint` loses the structural-match property (`const` patterns); non-riscv targets are unaffected.
- `const_eq`/`const_is_zero` are unchanged.

### DIFF
--- a/src/cmp.rs
+++ b/src/cmp.rs
@@ -43,19 +43,15 @@ impl<const BITS: usize, const LIMBS: usize> Ord for Uint<BITS, LIMBS> {
     }
 }
 
+// On riscv the derived `PartialEq` (a `[u64; LIMBS]` array comparison) lowers
+// to a `bcmp`/`memcmp` libcall, whose call overhead dwarfs the comparison
+// itself — these targets have no inline-memcmp expansion. Open-code a
+// branchless limb compare there; all other targets keep the derive, which LLVM
+// already lowers optimally.
+#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
 impl<const BITS: usize, const LIMBS: usize> PartialEq for Uint<BITS, LIMBS> {
     #[inline]
     fn eq(&self, other: &Self) -> bool {
-        as_primitives!(self, other; {
-            u64(x, y) => return x == y,
-            u128(x, y) => return x == y,
-        });
-        // Open-coded rather than derived: deriving `PartialEq` compares the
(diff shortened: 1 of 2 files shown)


DaniPopes:
I would rather gate this behind specific arch's where you measured. Rust lowers the slice equality to LLVM memcmp, which I would trust more than hand coding a vector equality. If there was a way to guarantee that it lowers to vector equality then I would be more in favor of that (`icmp eq <N x i64>` in LLVM IR)

Qumeric:
Done in 81d081f — gated to riscv, everything else keeps the derive. One caveat: measured on riscv32 (zkVM); I included riscv64 since the backend lowering is the same, but happy to narrow the gate if you'd rather keep it strictly to measured targets.

Qumeric:
^ this codspeed report looks like false positive. I think branch needed to be rebased on top of latest main

DaniPopes:
Yes it's flaky

Qumeric:
FYI: A relevant RISC-V optimization was recently enabled in LLVM: https://github.com/llvm/llvm-project/pull/209738

However, it will take around a year to land in stable Rust, so this change can still be beneficial in the time being.

DaniPopes (approved):

PR #611, opened 2026-07-08
docs: fix incorrect comments in add/mul/gcd algorithms

prestwich:
Fixes three documentation-only defects where comments/formulas contradicted the (correct) code.

| Location | Before | After |
| --- | --- | --- |
| `src/algorithms/add.rs:48` | `borrowing_sub_n` doc: `lhs -= rhs - borrow` | `lhs -= rhs + borrow` — matches per-limb `lhs[i] - rhs[i] - borrow` |
| `src/algorithms/mul.rs:168` | `submul_nx1` borrow = `⌊(a·b − lhs)/2^(64N)⌋` | `⌈(a·b − lhs)/2^(64N)⌉` — e.g. `lhs=[3]`, `a=[5]`, `b=1` returns borrow `1`, floor gives `0` |
| `src/algorithms/gcd/matrix.rs:259,262,275,288` | `(odd)`/`(even)` labels swapped vs. the returned parity flag | labels aligned (flag `false` = even index) |

No behavioral changes — comments and one KaTeX formula only. The Jebelean stopping conditions themselves are unchanged.

Docs verified to build without warnings:
```
RUSTDOCFLAGS="--cfg docsrs --html-in-header .cargo/katex-header.html -D warnings" \
  cargo +nightly doc --no-deps --all-features
```

### DIFF
--- a/src/algorithms/add.rs
+++ b/src/algorithms/add.rs
@@ -45,7 +45,7 @@ pub fn carrying_add_n(lhs: &mut [u64], rhs: &[u64], mut carry: bool) -> bool {
     carry
 }
 
-/// ⚠️ `lhs -= rhs - borrow`
+/// ⚠️ `lhs -= rhs + borrow`
 #[doc = crate::algorithms::unstable_warning!()]
 #[inline(always)]
 pub fn borrowing_sub_n(lhs: &mut [u64], rhs: &[u64], mut borrow: bool) -> bool {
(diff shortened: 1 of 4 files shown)


src/algorithms/add.rs
```diff
@@ -45,7 +45,7 @@ pub fn carrying_add_n(lhs: &mut [u64], rhs: &[u64], mut carry: bool) -> bool {
     carry
 }
 
-/// ⚠️ `lhs -= rhs - borrow`
+/// ⚠️ `lhs -= rhs + borrow`
```
DaniPopes, line 48:
```suggestion
/// ⚠️ `lhs = lhs - rhs - borrow`
```

DaniPopes (approved):
ig

PR #613, opened 2026-07-09
feat: add Uint::to_string

DaniPopes:
Adds an alloc-gated inherent `Uint::to_string` that intentionally shadows `ToString`. Decimal formatting now uses one shared routine that writes ASCII digits directly into either the existing stack buffer for `Display` or a preallocated byte vector for `to_string`; the vector capacity comes from a fixed-point upper bound for log10(2), avoiding compile-time big-integer division.

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -135,6 +135,21 @@ impl_fmt_pow2!(fmt::Octal; base::Octal, false);
 impl_fmt_pow2!(fmt::LowerHex; base::Hexadecimal, false);
 impl_fmt_pow2!(fmt::UpperHex; base::Hexadecimal, true);
 
+#[cfg(feature = "alloc")]
+impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS> {
+    /// Converts this integer to a decimal string.
+    ///
+    /// This method intentionally shadows [`ToString::to_string`].
+    #[allow(clippy::inherent_to_string_shadow_display)]
+    #[inline]
+    #[must_use]
+    pub fn to_string(&self) -> alloc::string::String {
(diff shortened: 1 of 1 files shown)


src/fmt.rs
```diff
@@ -169,18 +217,31 @@ impl<const SIZE: usize> StackString<SIZE> {
         unsafe { self.buf.as_mut_ptr().add(self.len).cast::<u8>().write(b) };
         self.len += 1;
     }
+
+    #[inline]
+    fn push_bytes(&mut self, bytes: &[u8]) {
+        debug_assert!(self.len + bytes.len() <= SIZE);
+        unsafe {
+            let dst = self.buf.as_mut_ptr().add(self.len).cast();
+            core::ptr::copy_nonoverlapping(bytes.as_ptr(), dst, bytes.len());
```
prestwich, line 226: ignore previous deleted comment. not UB when doing `self.push_bytes(self)`, but _is_ UB if there's an overflow on len, right? 🤔

src/fmt.rs
```diff
@@ -51,34 +51,15 @@ mod base {
 }
 use base::Base;
 
-macro_rules! impl_fmt {
-    ($tr:path; $base:ty, $base_char:literal) => {
-        impl<const BITS: usize, const LIMBS: usize> $tr for Uint<BITS, LIMBS> {
-            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                if let Ok(small) = u64::try_from(self) {
```
prestwich, line 58: fast-paths for u64/u128 intentionally removed?

DaniPopes:
Sorry need to re-do this, this is dogshit

PR #637, opened 2026-09-18
chore: pin optional dev dependencies and remove issue tracker

decofe:
Pin all 36 optional dependencies to exact patch versions in dev-dependencies, and require a pin whenever an optional dependency is added. Remove the unused Python issue-tracker workflow and script.

Validated native library tests with all features and without default features, wasm test compilation, dependency resolution, and formatting.

Prepared with AI assistance.

Prompted by: @grandizzy


.github/workflows/bench.yml
```diff
@@ -12,16 +12,23 @@ env:
...
         with:
           cache-on-failure: true
       - name: Install cargo-codspeed
-        uses: taiki-e/install-action@v2.87.11
+        uses: tempoxyz/gh-actions/vendor/taiki-e/install-action@25cce154e7fb10f99361a166468a6c56b9c31aa3
         with:
-          tool: cargo-codspeed
+          tool: cargo-codspeed@5.0.2
+          checksum: true
+          fallback: cargo-binstall
```
mablr: `cargo-codspeed` has no manifest in this pinned installer, so this always takes the [cargo-binstall fallback](https://github.com/tempoxyz/gh-actions/blob/25cce154e7fb10f99361a166468a6c56b9c31aa3/vendor/taiki-e/install-action/main.sh#L1093-L1107). `checksum: true` verifies the cargo-binstall bootstrap, but the fallback does not require a checksum/signature for the final tool binary. Please use an exact-version locked source install through Aegis, or add verified artifact support in `gh-actions` and disable fallback.

mablr, line 31: `cargo-codspeed` has no manifest in this pinned installer, so this always takes the cargo-binstall fallback https://github.com/tempoxyz/gh-actions/blob/25cce154e7fb10f99361a166468a6c56b9c31aa3/vendor/taiki-e/install-action/main.sh#L1093-L1107. `checksum: true` verifies the cargo-binstall bootstrap, but the fallback does not require a checksum/signature for the final tool binary. We should probably use an exact-version locked source install through Aegis, or add verified artifact support in `gh-actions` and disable fallback.

prestwich:
what is the "Aegis rollout"?

grandizzy:
> what is the "Aegis rollout"?

Aegis is a socket-dev API client, meant to intercept cargo (and other package managers) and to enforce cooldown

prestwich:
all deps are features. we provide API compatibility at specific majors for feature deps, and never need to update. simply pinning each one to a known-good version should be sufficient, no?

more generally, the goal here to protect our development process from supply chain attacks, right?

grandizzy:
indeed, exact pins would reduce version drift for cargo deps, though CI also installs Rust tools and other Python packages (these can also be automatically checked through Aegis) - could add the cargo lock and add secure runner only for the other jobs, would that work?

prestwich:
can you explain your threat model? what exactly are you trying to prevent? As a library, we can't tightly constrain dependency versions outside our own builds. Are you trying to prevent github/codspeed tokens from getting stolen?

grandizzy:
correct, the goal is to prevent malicious code from executing on the runner, where it could exfiltrate tokens (or other sensitive data)

prestwich:
the repo doesn't interact with any sensitive data or handle secrets, and the damage from the github token or codspeed token is already pretty well-mitigated. Do you have a specific threat model in mind?

I'd be interested in removing the python based issue tracking, as we don't really use it or maintain the issues it creates. It's leftover from the previous maintainer

grandizzy:
Cargo.lock + remove python should work, you up to PR it or I should repurpose this one?

prestwich:
Go ahead and repurpose. 

Using a lockfile would require manual gitignore finagling whenever a new feature is added. Rather than adding a `Cargo.lock`, ensure that all optional deps have pinned patches in the dev-dependencies section of `Cargo.toml`, and add a note to that block specifying that this MUST occur whenever a new optional dependency is added.

